PR #3, opened 2020-03-10
Linting setup

vinnyjth:


conrad-vanl:
@vinnyjth is it possible to move this linter config into an `@apollosproject` package?

redreceipt:
works good! I've got a new `yarn.lock` in the API folder. Also, it's a little frustrating that I can `yarn <COMMAND>` from the root on almost everything except just the generic `yarn`.

vinnyjth:
@conrad-vanl We could, do we want to prioritize that at this point? I don't know how much I care about controlling the specifics of the linter rules (I'm fine dropping a file in and let the churches have at)

@redreceipt Working on it :)

redreceipt:
This seems like a good time, I still think there could be something to stripping out all lint, test, storybook, and other unnecessary tools from the templates. Let churches decide all that and maybe have a doc on best practices.

vinnyjth:
@conrad-vanl @redreceipt What's our appetite on merging this vs waiting to build an apollo-lint-rules package?

redreceipt:
I think that's the way to go. Test package too :)

conrad-vanl:
@vinnyjth If it's easy to throw our standard eslint into a package that can be used like a plugin or something on church apps, that would be great. Less code that's subject to upgrade diffs.

vinnyjth:
@conrad-vanl  Done
https://github.com/ApollosProject/apollos-apps/pull/new/eslint-config-package

redreceipt:
that's awesome!

apollos-church-api/package.json
```diff
@@ -74,6 +75,14 @@
     "get-graphql-schema": "^2.1.2",
     "graphql-schema-linter": "^0.2.0",
     "nodemon": "^2.0.2",
-    "linkemon": "^0.1.1"
+    "linkemon": "^0.1.1",
+    "eslint": "^5.16.0",
+    "eslint-config-airbnb-base": "^14.0.0",
```
conrad-vanl: we still need all of these separate eslint packages? Won't the plugin take care of it?

apolloschurchapp/package.json
```diff
@@ -131,7 +132,20 @@
     "react-native-storybook-loader": "1.8.0",
     "react-native-version": "2.6.2",
     "react-test-renderer": "16.8.6",
-    "waait": "1.0.2"
+    "waait": "1.0.2",
+    "eslint": "^4.19.1",
+    "eslint-config-airbnb": "^17.0.0",
```
conrad-vanl: Same as above

apollos-church-api/package.json
vinnyjth: We _could_ but the official guidelines recommend you include them as peerDeps

vinnyjth: Well - I take that back a bit. We can bundle  `configs` and `parsers` but we need to include plugins as peerDeps. I can fix that in core but we'll have to ship another release first.

PR #7, opened 2020-03-12
Adds Node Version Config

redreceipt:


redreceipt:
Actually @vinnyjth I put this is premature, it seems Node 8 isn't compatible. What should we do here?

Screen Shot 2020-03-12 at 8 37 47 AM

vinnyjth:
@redreceipt Let's bump the version in the `.nvm` files to 10. That's the version the latest RN version recommends, and our app runs fine on either (8 or 10)

We'll probably want to bump the `engine` fields in the `package.json` as well.

redreceipt:
You think it's important to bump the `apps` repo too? I can do it if so.

vinnyjth:
Sure! Thanks

PR #13, opened 2020-03-17
Update README.md

redreceipt:


burkeshartsis:
Heads up, I think I need to tweak the CI settings because tests don't seem to run until _after_ a PR is created and you push. So if you push an arbitrary change to this PR it will trigger it to run but if you create a PR with all of your changes already made (like this one-liner) it wont ever run. I'll open a PR now to fix it.

PR #14, opened 2020-03-18
Use InteractWhenLoaded component and Renames ViewedContentx to ViewedContent

vinnyjth:


vinnyjth:
Companion PR: https://github.com/ApollosProject/apollos-apps/pull/1395

vinnyjth:
@conrad-vanl Good catch. I believe Willow fixed this in their repo and Newspring is using an alias to rename it in Amplitude.

PR #16, opened 2020-03-23
Fixes potential for client to not link all packages.

vinnyjth:


vinnyjth:
@redreceipt You have technically already tested this ;)

PR #22, opened 2020-03-30
Heroku  / AppCenter deployment setup

vinnyjth:
For both Heroku and App Center, we're back to being able to point those services at this repo and autodeploy. I have both AppCenter and Heroku pointing and deploying from this branch, if you want proof.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:8-alpine
+FROM node:10-alpine
 COPY . /usr/src/
 WORKDIR /usr/src
 RUN yarn
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,7 @@
+FROM node:8-alpine
+COPY . /usr/src/
+WORKDIR /usr/src
+RUN yarn
+WORKDIR ./packages/apollos-church-api
+EXPOSE 4000
+CMD [ "yarn", "start:prod" ]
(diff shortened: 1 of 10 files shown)


Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM node:10-alpine
+COPY . /usr/src/
+WORKDIR /usr/src
+RUN yarn
```
vinnyjth: for all our tooling

Dockerfile
```diff
@@ -0,0 +1,8 @@
...
+WORKDIR ./apollos-church-api
+RUN yarn
```
vinnyjth: for the package deps

apolloschurchapp/appcenter-pre-build.sh
```diff
@@ -6,7 +6,15 @@ cp $droid_dir/gradle.properties.appcenter $droid_dir/gradle.properties
...
 
-cd ../../..
-yarn
-yarn lerna run generate-stories
+# You can comment this out if your app doesn't have an existing build number.
+VERSION_CODE=$((VERSION_CODE_SHIFT + APPCENTER_BUILD_ID))
+
+echo Using "$VERSION_CODE" as build number
+
+echo $VERSION_CODE
```
vinnyjth: This is something that all our partners have needed a solution for. This lets us use both incremental build numbers in app center, and avoid running into the "build number must be greater than last build number" issue. (AppCenter always starts at 1 for incrementing build numbers)

PR #24, opened 2020-04-01
Adds HorizontalCardListFeature component

burkeshartsis:
Adds HorizontalCardListFeature component 💥 

Blocked by https://github.com/ApollosProject/apollos-apps/pull/1404


burkeshartsis:
Something worth considering. Much of this component is pretty similar to `HorizontalContentSeriesFeedConnected`. @conrad-vanl even proposed that with some work these could be the same component. Fundamentally, the differences lie in how the `ConnectedComponents` map data to their pure component children.

apolloschurchapp/src/tabs/home/Features/HorizontalCardListFeature/index.js
```diff
@@ -0,0 +1,82 @@
...
+
+const Header = styled(({ theme }) => ({
+  paddingTop: theme.sizing.baseUnit * 3,
+  paddingBottom: theme.sizing.baseUnit * 0.5,
+}))(PaddedView);
+
+class HorizontalCardListFeature extends PureComponent {
+  loadingStateObject = {
+    id: 'fakeId0',
+    isLoading: true,
```
conrad-vanl, line 33: weren't you and @richarddubay just talking about this?!

burkeshartsis:
> Is there a plan to bring tabs/home/Features into a ui- package? Doesn't seem like it belongs in this repo...

I completely agree it needs a better home. I'm hoping to accomplish that in another PR. I would LOVE any Ideas on where you think these components should live. `ui-connected` is the easy choice because it already exists. However, that package is largely a 1-1 map of `ui-kit` to `ui-connected` components. It doesn't mean it has to stay that way though.

apolloschurchapp/src/tabs/home/Features/HorizontalCardListFeature/index.js
burkeshartsis: We were. That's one  of the reasons I was so familiar with what he was trying to do. The code was very front of mind. I paired with him for a bit and got a few take aways we can discuss later on how we can improve some of the code he was working with.

PR #28, opened 2020-04-09
Fix cache-lookup on content-single initial open

conrad-vanl:
So at some point our `getContentItem` query inside of `content-single` was referencing our connected content card query. The whole reason that we have a query tied to each `content-card` is so that we can use the same fragment when opening a content card so that we can show content immediately on render, instead of flashing the loading state. Apollo GraphQL requires ALL of the data you're querying for to exist in the cache. The easiest way we can guarantee that is by using the same fragment.

TL;DR: With this PR you'll no longer see a flash of loading state when tapping on a content card.


vinnyjth:
I agree with this PR. Do you think we should remove the `CONTENT_ITEM` fragment long term?

apolloschurchapp/src/content-single/getContentItem.js
```diff
@@ -10,5 +10,5 @@ export default gql`
       }
```
vinnyjth, line 10: You'll need to rename the fragment referenced on line 9 to `contentCardFragment`

conrad-vanl:
@vinnyjth perhaps? I'm not sure. Maybe we should drop `contentCardFragment` and `contentItemFragment` and just make like a "basic" or "minimal" contentFragment of some sort.

vinnyjth:
> Maybe we should drop contentCardFragment and contentItemFragment and just make like a "basic" or "minimal" contentFragment of some sort.

This might be our best option. Merging this as is for now.

PR #29, opened 2020-04-09
Adds LiveConsumer support to Vertical/Horizontal card list

vinnyjth:
This brings us to feature parity with the content card connected. Plus more! Huzzah! 

I swapped the default feed around a bit so you'll be able to see the "Spur One Another" item go live. It goes live at the bottom half of every hour.


vinnyjth:
I would love to have tests for this, but we don't have a good way of testing the live consumer yet :/

vinnyjth:
This PR will now need to be ported to core.

PR #35, opened 2020-04-11
Adds link-doctor script and other improvements. 

vinnyjth:
A few things in this PR 

- Adds `watchman watch-del-all` to `yarn start` which fixes a number of linking issues.
- Adds `yarn link-doctor` which should help debug some workspace issues. It won't catch 100% of issues at this point, so it's very helpful if you continue to point out places where syncing isn't working @redreceipt @conrad-vanl @jordanwade @burkeshartsis. The more I know about what the issues are, and the more info I can collect, the better we'll be able to improve the `doctor` and linking in general. 
- Converts to a `nohoisted` yarn workspace so you can `yarn` from root like ya'll wanted.


burkeshartsis:
> Converts to a nohoisted yarn workspace so you can yarn from root like ya'll wanted.

🙌🙌🙌

vinnyjth:
I need to cut a release before this will pass. Will do that as soon as the shadow PR merges.

PR #47, opened 2020-04-28
Makes PORT variable optional

redreceipt:


vinnyjth:
At one point I think we had this a specific way for tests that were importing files from the API. I believe that is no longer the case, but maybe run the tests to confirm.

redreceipt:
There are no tests in the API template

—
Michael

vinnyjth:
> There are no tests in the API template

It was the app tests that were importing from the API. I think we fixed that though.

redreceipt:
I'm sorry I'm confused. Not sure what I need to do here.

PR #64, opened 2020-06-15
Adds blur package

redreceipt:


vinnyjth:
Sounds like this package is going away real soon, I think we are waiting for a PR in core to get merged. For now, just plan on adding this package to your template if you are working on an unrelated PR.

/cc @burkeshartsis

burkeshartsis:
This is correct! Move fast and break things or something like that

PR #71, opened 2020-07-02
Only onboard once

christinamegura:
This PR adds a feature to only show onboarding the first time you sign in to a new device by storing if a user has finished onboarding and checking onboarding status before opening onboarding.


conrad-vanl:
@vinnyjth I'd like you to review this and handle getting it merged in, thanks!

apolloschurchapp/src/ui/Onboarding/index.js
```diff
@@ -57,43 +59,48 @@ function Onboarding({ navigation }) {
             <LocationFinderConnected
               onPressPrimary={swipeForward}
               onNavigate={() => {
-                navigation.navigate('Location');
+                navigation.navigate('Location', { onFinished: swipeForward });
```
vinnyjth, line 62: We should probably back this change out :) It's specific to Willow.

apolloschurchapp/src/ui/Onboarding/index.js
```diff
@@ -57,43 +59,48 @@ function Onboarding({ navigation }) {
...
-                  if (checkRes.status === RESULTS.DENIED) {
-                    requestNotifications(['alert', 'badge', 'sound']).then(
-                      () => {
-                        update();
+            <Query query={WITH_USER_ID} fetchPolicy="network-only">
+              {({
+                data: { currentUser: { id } = { currentUser: { id: null } } },
+              }) => (
+                <AskNotificationsConnected
+                  description={
```
vinnyjth, line 75: This also is specific to Willow, and can be removed :)

apolloschurchapp/src/ui/Onboarding/index.js
```diff
@@ -57,43 +59,48 @@ function Onboarding({ navigation }) {
...
+                    'Get updates, reminders, and announcements from your Church.'
+                  }
+                  onPressPrimary={() => {
+                    onboardingComplete({ userId: id });
+                    navigation.replace('Tabs');
```
vinnyjth, line 80: We should keep the `dispatch(NS.resetAction)` that was here before. Willow should be updated to use the `dispatch(NS.resetAction)` eventually :) `.replace()` lets you hit back and return to the onboarding screens after logging in)

apolloschurchapp/src/Providers.js
```diff
@@ -19,7 +20,12 @@ const AppProviders = (props) => (
       <AuthProvider
         navigateToAuth={() => NavigationService.navigate('Auth')}
         navigate={NavigationService.navigate}
-        closeAuth={() => NavigationService.navigate('Onboarding')}
+        closeAuth={() =>
```
vinnyjth, line 23: Perfect!

PR #90, opened 2020-08-24
Adds Fastlane and Certificates

redreceipt:
Allows Fastlane deploys to beta for iOS and Android and automated certificate management.

This one is in a weird spot. I can't test the deploys unless I'm on the master branch so it may require a little experimentation after merge to get 100%. But it's close.

To pull certs, run `bundle exec fastlane ios certificates` from app directory


redreceipt:
This one was a lot bigger than I originally anticipated and it's still not 100% but I think this is a good first go at it. I tried to stick to `fastlane init` as best I could in most circumstances and fill in the gaps that weren't working great with notes from various tutorials.

apolloschurchapp/fastlane/Fastfile
```diff
@@ -0,0 +1,43 @@
...
+platform :ios do
+  desc 'Fetch certificates and provisioning profiles'
+  lane :certificates do
+    match(app_identifier: 'com.differential.apollos', type: 'development', readonly: true)
+    match(app_identifier: 'com.differential.apollos', type: 'appstore', readonly: true)
+  end
+
+  desc "Push a new beta build to TestFlight"
+  lane :beta do
+    increment_build_number(xcodeproj: "apolloschurchapp.xcodeproj")
```
vinnyjth, line 31: we might need to rethink this bit. If we increment the build number, but don't commit and push, we'll end up incrementing to the same value every time. We might need a way to look at the last published build number, or use the number of commits in the repo, or something like that.

redreceipt: for newspring we did it by date. I can have fastlane actually do a commit and push. for now I just wanna test the manual deploy.

vinnyjth (approved):

PR #91, opened 2020-08-24
Adds Canary Packages to Dev Process

redreceipt:
run `yarn canary` or `yarn latest` from either the API or App directories and it will update to all canary builds or all stable builds. This is so we can go back to only having a master branch and using the latest code when necessary to build the app from NPM.

Screen Shot 2020-09-04 at 1 57 46 PM


redreceipt:
@vinnyjth I worked some bash magic, what do you think about this approach?

redreceipt:
@vinnyjth good morning! back from baby vaca :) this one is ready for a second look

vinnyjth (commented):
Looking good! Two more pieces of feedback.

1. Can we add the commands to run this from the root? Something as simple as `"canary": "yarn lerna run canary"
2. I'm getting errors when trying to switch the API related to the `postinstall` hook. Looks like it's trying to build before the babel package is installed. Seems to work fine if you add `--ignore-scripts` to the ends of the `yarn add`s.

redreceipt:
what do you think about moving `yarn build` to only happen with `start:prod` instead of `postinstall`? Seems more specific and appropriate.

vinnyjth:
We don't want to build before every start :/ so it's kinda tricky. We need like a `post-prod-install` more than anything.

vinnyjth:
In other words - I think having `yarn build` in the `postinstall` is the right place.

redreceipt:
Gotcha. I just hate using `ignore-scripts` because somehow we'll have to remember to `build` after adding the packages. Trying to figure out a way to get babel in there early or something.

Also, running commands from the root gives me all sorts of trouble. I'll keep hacking on it though for a bit.

vinnyjth:
> Also, running commands from the root gives me all sorts of trouble. I'll keep hacking on it though for a bit.

Interesting - have you tried adding `ignore-scripts` first?

redreceipt:
Honestly I don't understand why this error in the first place. It should already be there...

`{ Error: Cannot find module '@apollosproject/babel-preset-apollos' from '/Users/michael/Documents/Projects/apollos/apollos-templates/apollos-church-api'`

vinnyjth:
yah that's the same error I got before I added --ignore-scripts. I think it has to do with adding all the packages at once, but I'm honestly not sure.

redreceipt:
Ok so adding `ignore-scripts` works from each of the directories but running from root still bombs out.

Screen Shot 2020-09-09 at 10 17 14 AM

vinnyjth:
That error makes sense - we don't actually have a canary version for that package since we cloned it and made a few changes to it. Not sure the best way to fix, might need a flag to ignore certain errors?

redreceipt:
Can we move forward by not having root commands and solving that in another PR? Worse case is someone developing needs to build the template app and hits an error, they then run `yarn canary` from both dirs and voila. Confident a good solution is possible, just low priority.

vinnyjth (approved):
👍

PR #98, opened 2020-09-10
Fixes Heroku Docker Deploys

redreceipt:


vinnyjth:
Thanks for doing this, I think I've had to make the same change in client apps for the last couple of new clients.

vinnyjth (approved):

PR #104, opened 2020-09-14
Updates to new church online enpoint

redreceipt:


redreceipt:
blocked until new church online code is released

vinnyjth (approved):
I believe this is GTG now

redreceipt:
i'd rather wait until it's in a release so we can point master to it

redreceipt:
it's just to get the API to stop throwing errors on heroku

vinnyjth:
How about we get it into `develop` so we can have the change locally?

vinnyjth:
It will end up `master` automatically post release that way

PR #111, opened 2020-09-17
Splits CI

redreceipt:
This is just for better visibility on the PR status page


redreceipt:
@vinnyjth in case you were wondering, this is the one that's blocking all the rest

vinnyjth (approved):
Love it, I was going to suggest you do this :)

PR #107, opened 2020-09-15
NodeSingle route and linking fixes

vinnyjth:
**IMPORTANT: Unless you are doing a release, you should be pointing your PR at the `develop` branch. `master` should not rely on code not published in `npm`**

To test, link with `apps` and then visit the home feed. Tap on the "View More" and a content item should open up. This item opening is the NodeSingle, rather than the usual ContentSingle.

PR #116, opened 2020-09-17
Updates to Canary Builds

redreceipt:


redreceipt:
doesn't necessarily need to be up to date, if a dev wants another dev to use it to test, they'll create a PR against canary to update one or all packages

vinnyjth:
>  if a dev wants another dev to use it to test, they'll create a PR against canary to update one or all packages

I don't think this will work unless the code that the dev is testing has already been merged, right? We still have our chicken/egg situation :/ 

I see the value in this, I'm just not sure if it will end up helping at all.

vinnyjth:
Oh, I remember the other problem with doing this, it makes it hard to merge into `master` :/ you have to go and update all your package versions (which we already do when going from `develop -> master` today I suppose, but in that situation we're updating the versions  to reflect the code becoming mature)

redreceipt:
Ok, I see. We can still use this flow but have to be mindful of breaking changes. I think this will be better in the long run anyway. This is the flow I'm thinking tell me where it fails:

*Breaking*

- Vincent is adding breaking features to apps (has new native dependency)
- Vincent creates PR in apps against `next`
- Vincent creates a PR in templates and merges it into `next` with a new iOS pod
- Michael wants to test the new feature, checks out `next` on templates
- Michael approves PR in apps

*Non-breaking*

- Vincent is adding new non-breaking features to apps
- Vincent creates PR in apps
- Vincent creates a branch in templates to use for testing
- Michael wants to test the new feature, checks out the feature branch
- Michael approves PR in apps
- Vincent creates PR in templates with feature branch against canary

*Update master*

- Once new features are released to stable, Create PR from `next` or `canary` to `master` updating version numbers

redreceipt:
I modified the above comment slightly, re-read if you see this on email :)

redreceipt:
If you like that flow and it makes sense, I'll document that as part of the development README in apps repo

vinnyjth:
@conrad-vanl I'm mentioning you here since you have good feedback on flows like this ^

vinnyjth:
@redreceipt I'm still wrapping my head around this. I think it makes since, except maybe we need to define `breaking changes` better. Once I swapped "breaking changes" for "2.0 work" the flow made more sense to me.

redreceipt:
> @redreceipt I'm still wrapping my head around this. I think it makes since, except maybe we need to define `breaking changes` better. Once I swapped "breaking changes" for "2.0 work" the flow made more sense to me.

Yes. Except breaking changes in apps has to include new packages like `react-native-airplay` that require a new iOS pod. If 1.6.0 had that in it, it would break apps, so that would have to be in 2.0 code. Maybe we're saying the same thing.

redreceipt:
it's important to think of canary and next branches in templates as strictly testing branches. and then when merging back into master, the PR just has to change all the version numbers back.

redreceipt:
Shouldn't need to use `canary` all that often anymore because of the storybook update in apps

vinnyjth:
> Yes. Except breaking changes in apps has to include new packages like react-native-airplay that require a new iOS pod. If 1.6.0 had that in it, it would break apps, so that would have to be in 2.0 code. Maybe we're saying the same thing.

In that case, would pretty much any change in `templates` count as breaking?

vinnyjth:
Here's an example of a situation that I'm not sure how to handle in the new flow: 

https://github.com/ApollosProject/apollos-templates/pull/107

It's not _breaking_ since you don't need it for your app to function, but it sets you up to use enhancements that will eventually go live in `apps`, but aren't live yet. Maybe it would help to understand how you would handle a PR like this.

redreceipt:
> > Yes. Except breaking changes in apps has to include new packages like react-native-airplay that require a new iOS pod. If 1.6.0 had that in it, it would break apps, so that would have to be in 2.0 code. Maybe we're saying the same thing.
> 
> In that case, would pretty much any change in `templates` count as breaking?

Any change to templates master branch is _already_ breaking. But templates isn't a dependency so I don't think we need to think of updates to it as breaking changes. it's version-less. It should always be assumed the patch tool is going to patch breaking changes and should be done carefully.

redreceipt:
> Here's an example of a situation that I'm not sure how to handle in the new flow:
> 
> #107
> 
> It's not _breaking_ since you don't need it for your app to function, but it sets you up to use enhancements that will eventually go live in `apps`, but aren't live yet. Maybe it would help to understand how you would handle a PR like this.

That PR would just go into master. The patch tool ignores config.yml ideally anyway.

vinnyjth:
Just to clarify, `master` in apps, but `canary` in templates right?

redreceipt:
> Just to clarify, `master` in apps, but `canary` in templates right?

Ah, yes. It would be merged into canary (templates). Which could not be updated to master(templates) until that code is released on a stable version.

redreceipt:
> Are you essentially recommending that above, except generalizing v2.0 as `next` and renaming `develop` canary?

yes :) and waiting to update master until stable releases are out

redreceipt:
but also, it's important to note, ALL template code does not need to go into canary unless it depends on a canary release.

vinnyjth (approved):

PR #121, opened 2020-09-18
Removes Root URL from Config

redreceipt:
doesn't look like we need it

### DIFF
--- a/apollos-church-api/config.yml
+++ b/apollos-church-api/config.yml
@@ -11,7 +11,6 @@ APP:
   # This variable is used for generating deep links.
   # It should match whatver is in your Info.plist `CFBundleURLTypes`
   DEEP_LINK_HOST: apolloschurchapp
-  ROOT_API_URL: ${ROOT_URL}
   JOBS_USERNAME: ${JOBS_USERNAME}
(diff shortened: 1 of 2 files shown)


vinnyjth:
https://github.com/ApollosProject/apollos-apps/blob/master/packages/apollos-data-connector-passes/src/resolver.js#L66-L67

redreceipt:
haha dang...

vinnyjth:
Yahhh... I would advocate for consolidating this field or/and adding better comments to this whole file.

redreceipt:
> Yahhh... I would advocate for consolidating this field or/and adding better comments to this whole file.

Explain what you mean. I feel like there should be a way to pull that base URL built into Apollo Server...

vinnyjth:
@redreceipt I was assuming that there were a few more parts of our API that use a similar field (like universal linking) but a glance, I guess not. 

Maybe we could rename this field to be more specific to the `PASS` feature?

vinnyjth:
There's not a default way to get an app's domain name from inside Heroku, if that's what you were thinking.

redreceipt:
yea I like that

PR #125, opened 2020-09-18
Rn63

vinnyjth:

@redreceipt Did I do this right? (am I pointed at the right branch?)


redreceipt:
Ok so this one is deceiving because it seems like such a big deal but I think it's perfect example of a PR that could be merged directly into master. Remember, templates is versionless so as long as it won't break when merged into master (meaning it doesn't rely on unreleased code) I think we put it straight into the repo. It's purely app template specific and doesn't care about any npm dependencies.

vinnyjth:
> I think we put it straight into the repo. It's purely app template specific and doesn't care about any npm dependencies.

So I totally agree with you. However... merging this into master would block existing churches from updating to new code that we introduce in templates (like upcoming node interfaces stuff) without also needing to update to RN 63 (which will be a pretty intense upgrade)

If it was just us, I would put in master and call it a day. However, for the sake of the client churches I want to make sure that we can continue to release 1.6, potentially 1.7 and so forth without requiring them to labor through an update that requires merging a lot of yucky conflicts.

redreceipt:
In that case I think then yes. We put it in next because it’s something we aren’t ready to offer to churches yet.

vinnyjth:
This is booting, but requires a linked setup until this pr is merged: https://github.com/ApollosProject/apollos-apps/pull/1579

redreceipt:
now that beta.0 is in here, should be good to test. I'm excited about this one

redreceipt:
getting a pod install error
Screen Shot 2020-09-22 at 2 46 09 PM

vinnyjth:
Nuke your pods folder.

redreceipt:
weird. I nuked Pods only and the yarn.lock changed...

vinnyjth:
Lol, stuff like that has happened a ton for me this week. Are things working for ya yet?

redreceipt:
oh yea 💪 
Screen Shot 2020-09-22 at 3 01 33 PM

redreceipt:
actually maybe not
Screen Shot 2020-09-22 at 3 03 18 PM

vinnyjth:
Mentioned this in slack - need to pull in canary builds that contain this fix: https://github.com/ApollosProject/apollos-apps/pull/1579

redreceipt:
is this one ready for me to officially review by the way?

vinnyjth:
@redreceipt Yah, since we got the the canary versions in there.

redreceipt:
android is failing for me 
Screen Shot 2020-09-22 at 3 52 03 PM

vinnyjth:
@redreceipt Thanks. We probably need to bump our maps package version. I'll take a look.

apolloschurchapp/scripts/add-packages.sh
```diff
@@ -26,5 +26,8 @@ else
 		PKGS=$(echo "$JSON" | sed -n "$DEPSLINE",/^$/p | grep "@apollosproject" | tr -d "\n")
 		DEVPKGS=$(echo "$JSON" | sed -n "$DEVDEPSLINE","$DEPSLINE"p | grep "@apollosproject" | tr -d "\n")
 fi
+
+PKGS=$(echo "$PKGS" | sed -e "s/@apollosproject\/react-native-airplay-btn@canary//")
```
redreceipt, line 30: this would need to account for beta and latest as well, I'm fine just running from each dir and circling back around on this feature if you're cool with it.

apolloschurchapp/.gitignore
```diff
@@ -133,7 +133,6 @@ DerivedData
 *.hmap
 *.ipa
 *.xcuserstate
-project.xcworkspace
```
redreceipt, line 136: can you think of why this is included in the repo now?

vinnyjth: I don't know, but it's part of the upgrade.

vinnyjth:
@redreceipt Was able to build locally with my latest commit.

redreceipt:
yuh
Screen Shot 2020-09-22 at 4 38 58 PM

redreceipt (approved):

PR #130, opened 2020-09-21
Disable fetch cache

vinnyjth:

If we want caching, we should use Fastly, Apollo's in memory cache, or a server side caching solution. The built in fetch cache is one too many variables to maintain firm control, over what ends up in the cache.


redreceipt:
this should go to master, easy switch for churches, no dependencies

vinnyjth:
@redreceipt done

redreceipt (approved):

PR #131, opened 2020-09-21
Adds environment variables

redreceipt:


redreceipt:
@vinnyjth can I get a thumbs up on this one?

vinnyjth (approved):

PR #132, opened 2020-09-21
v1.6.0-beta.0

redreceipt:


redreceipt:
we should catch develop up to master and then delete it. but first, we'll need to do a stable release with all the new "Node" stuff in it

vinnyjth:
So the "correct" way to do this is in a release PR :)

redreceipt:
> So the "correct" way to do this is in a release PR :)

I figured. that's why I marked it blocked. Can't merge this until we have a stable 1.6.0 to point master to. That's still the part of our whole process I'm fuzzy on because I haven't walked through it myself yet.

redreceipt:
@vinnyjth are you expecting me to just merge this in?

vinnyjth:
Yep! After tests pass, go for it.

PR #138, opened 2020-09-22
Master > Next

redreceipt:


vinnyjth (approved):

vinnyjth:
I think you're fine to merge this with or without tests passing (but I'm curious if they pass)

redreceipt:
> I think you're fine to merge this with or without tests passing (but I'm curious if they pass)

me too

redreceipt:
well would you look at that...

vinnyjth:
I feel like we should get a badge for that

PR #146, opened 2020-09-23
Adds Splash and Icon Generation Scripts

redreceipt:
MAJOR WIN! This allows us to automatically generate icons and splash screens with a command:

```
yarn icons <file>
yarn splash <file>
```

I had to move us from `react-native-bootsplash` to `react-native-splash-screen`


vinnyjth:
I really hate to say this, and I don't think this invalidates this PR, but we technically have support for this already with `yarn generate-bootsplash`. I think it works a little differently, it pulls from that `assets` folder, but it has the advantage of not needing further native code changes to get it working.

redreceipt:
Yea I don't love the number of native code changes but with this package, we get icons and splash. plus the splash generation with this package is WAY easier. The former has you answer a bunch of questions and you have to name the file correctly.

redreceipt:
plus no assets folder required

vinnyjth:
That's enough to sell me @redreceipt, but I vote we point this at the `next` branch. I would rather slip all our native changes together.

redreceipt:
> plus no assets folder required

for images that is

redreceipt:
moved it to `next`. what do you think the timeline on 2.0 is? like do you think we'll onboard any more churches with 1.x? If so, I'll need to document how to generate icons and splash the old fashioned way for those churches.

vinnyjth:
@redreceipt Conrad and I talked about this for allittle bit. It's probably something we'll wrap up for the Holidays (so I'm assuming not the next cycle, but the one after). It's hard to say at this point if we will onboard any churches onto 1.x though, at this point I think we could probably count on onboarding new churches onto an beta 2.x build.

vinnyjth:
@redreceipt for our current version, you could link out to the official docs: 

https://github.com/zoontek/react-native-bootsplash#setup

redreceipt:
@vinnyjth still good with this one?

vinnyjth:
@redreceipt yah. looks good to me. I haven't tested but I'll approve

vinnyjth (approved):

PR #149, opened 2020-09-24
Turns off Storybook on Production

redreceipt:
Not sure how this is working on iOS production but this is an attempt to get Android to build on the CI


redreceipt:
actually this doesn't make sense why this would help

PR #152, opened 2020-09-24
Publishes Production Schema from CI

redreceipt:


redreceipt:
need to provide GET query to get back a 200 somehow on the healthcheck

apollos-church-api/src/server.js
```diff
@@ -60,6 +60,11 @@ const apolloServer = new ApolloServer({
 
 const app = express();
 
+// health check
+app.get('/health', (req, res) => {
+  res.send('ok');
+});
+
```
redreceipt, line 67: @vinnyjth what do you think about this? simple and elegant. and I constantly wanted something like this at newspring. main purpose is to validate the deploy.

redreceipt: @vinnyjth you see this?

vinnyjth: Thanks for the ping, this is good!

vinnyjth (approved):

PR #156, opened 2020-09-24
Removes Unnecessary Tools

redreceipt:


apolloschurchapp/app.json
```diff
@@ -1,4 +0,0 @@
-{
```
vinnyjth, line 1: I think we still need this file.

vinnyjth: https://github.com/react-native-community/cli/issues/1113

apolloschurchapp/appcenter-post-clone.sh
```diff
@@ -1,16 +0,0 @@
-#!/usr/bin/env bash
```
vinnyjth, line 1: We should probably leave these in for now. I don't want them to be accidentally deleted from our client repos when they update and cause confusion and force a premature migration to Github Actions.

apolloschurchapp/package.json
```diff
@@ -12,16 +12,11 @@
     "setup": "./scripts/init.sh",
     "android": "SIM=\"$($HOME/Library/Android/sdk/emulator/emulator -list-avds | sed -n 1p)\" && $HOME/Library/Android/sdk/emulator/emulator -avd \"${SIM}\" & react-native run-android --appId com.differential.apollos",
     "codecov": "cat ./coverage/lcov.info | codecov",
-    "fixlint": "eslint ./src --fix",
```
vinnyjth, line 15: nice ;)

apolloschurchapp/app.json
redreceipt: don't think we do. it came from expo, was just to pull the name from inside the index.js file (which ours doesn't do)

apolloschurchapp/appcenter-post-clone.sh
redreceipt: fair. I could a separate PR to remove these pointed to next

apolloschurchapp/app.json
vinnyjth: I ran rn init recently, let me dig up what it included....

vinnyjth: okay - so the default react-native app does create that app.json.

redreceipt: ah. so according to that issue, it's created on init but not used anymore. I can put it back, it's just another file churches *should* rename even though it seems unnecessary

redreceipt: was trying to avoid as much renaming as possible

vinnyjth: that's fair - none of our churches have ever renamed it to my knowledge so I'm fine with deleting.

vinnyjth (approved):

PR #157, opened 2020-09-24
Removes ESLint

redreceipt:


redreceipt:
ah, nohoist is killing this...

redreceipt:
@vinnyjth is there a reason we are nohoist'ing everything or could we just do the packages we know cause problems?

vinnyjth:
@redreceipt The packages that cause problem list turns out to be very long. Easier to no hoist everything. Also stops us from having problems that don't surface when you're hoisting, but our clients run into when they aren't hoisting.

redreceipt:
gotcha, I suppose I could just put apollos/eslint config at the root

redreceipt:
nope that didn't work

PR #165, opened 2020-09-25
Adds Secrets to Repo

redreceipt:
This gives us a simple way to pass secrets and environment files back and forth. It requires a single password. I didn't want to mess with keys and all that, primarily because I don't understand it, but also because I don't think it's necessary for our purposes.

I think this will be a major win in helping us manage all the secrets for the different churches.


vinnyjth:
so cool looking I'm actually going to clone and test this one

redreceipt:
Ok, something I don't love, we really can't control if someone changes the password and uploads it. PR reviewers need to test the decrypt if the encrypted file is changed. Unlikely but it's a possibility.

vinnyjth:
Love this. A few thoughts. 

1. Can we make it more automatic? Thinking that `decrypt` should dump the `.env.enc` into the `.env` and decrypt should package the `.env` into `.enc.env`. Maybe keep the script out of the root and run it from each package folder? 

2. This person says that we shouldn't use aes-256-cbc and should use something else instead for security reasons: https://stackoverflow.com/a/16056298 I think he's probably right.

vinnyjth:
@redreceipt The nice thing is that if someone does change the password, we can always rollback because `git`. That's something we can't do right now in Heroku/AppCenter.

vinnyjth:
@redreceipt If we stick with `ssl` we should do something like this guy recommends if possible.

redreceipt:
> 1. Can we make it more automatic

Ok so yes. I thought about this a bit. The reason I left it open-ended is that we will probably only encrypt once when a repo gets set up. And it's not just `.env` files but Android and potentially iOS files later if we want. Maybe we would want to try and keep `.env` files in sync for people but I think that may get annoying if my env file changes all the time. 

All that being said if you feel strongly about making it a bash script and just hardcoding which files to encrypt and decrypt in that script, we could do that. Just may have to edit that script as the file list grows and shrinks.

Currently telling people to just `yarn decrypt <file>` in the setup README I figured would suffice for now.

vinnyjth:
> Currently telling people to just yarn decrypt <file> in the setup README I figured would suffice for now.

Fair enough. Let's start here and see how much pain is felt.

redreceipt:
i tried to run that guys command and it didn't work
Screen Shot 2020-09-25 at 2 25 48 PM

redreceipt:
also it seems if we make the number of iterations public (as he suggests) it kinda defeats the purpose. But I could be thinking about that wrong.

vinnyjth:
I like the idea of using this tool. 

https://github.com/FiloSottile/age

Requires brew install though, which I think is okay for something like this.

vinnyjth:
Example of using passphrases: https://github.com/FiloSottile/age#passphrases

redreceipt:
and actually, instead of encrypting the `.env` file, I think I'd rather encrypt an `.env.shared` file and just put in the variables that are necessary. For example I threw in a personal auth token for match without thinking and the `APPS_LOCATION` variable, which they don't need.

redreceipt:
openssl can do all that as well (file keys or rsa keys), I'd really like to stick with it since it's built in out of the box for macs, we just have to decide on the strategy and I'm confident I can make it work for us.

vinnyjth:
> and actually, instead of encrypting the .env file, I think I'd rather encrypt an .env.shared file and just put in the variables that are necessary. For example I threw in a personal auth token for match without thinking and the APPS_LOCATION variable, which they don't need. 

I like this idea. 

> openssl can do all that as well (file keys or rsa keys), I'd really like to stick with it since it's built in out of the box for macs, we just have to decide on the strategy and I'm confident I can make it work for us.

This might be tricky. Turns out the MacOS version of OpenSSL doesn't support `pkbdf2`. I'm not convinced that we can't use `openssl`, but it might be tricky.

redreceipt:
if you're confident that age tool is secure enough, let's just go with that. It does make me nervous having any question about openssl when we'll be storing rock admin api keys on the internet with access to tithing.

redreceipt:
Oooor looks like the most recent version of open SSL on home brew includes that flag so if we're going to ask for an external install might as well just use updated open SSL because default macOS hopefully will include it soon and I like the open source security aspect of it. What do you think?￼￼

redreceipt:
I bring it up because I'm nervous about getting age and go to work on GitHub actions￼

redreceipt:
alright @vinnyjth take a look. I looked at age but hit some roadblocks. for mvp we can just explicitly use openssl 1.1 after a brew install.

vinnyjth:
Tested locally, didn't even need to install anything! Looks good!

vinnyjth (approved):

PR #166, opened 2020-09-28
Removes False Keystore

redreceipt:
This keystore was my attempt to gen a new one but it didn't work so now I'm waiting on Google to see if they'll let me download it from them.


redreceipt:
actually i may be able to upload this one to google...

redreceipt:
actually just delete this one and we'll start from scratch if that's what they need me to do

vinnyjth (approved):

PR #184, opened 2020-09-30
Makes Releases Easier

redreceipt:
This one is cool. 

`yarn canary` now works from the root

`yarn bump` will look for the latest tag in `apps`, install packages based on tag ("beta" or "latest") and then bump the `apollos.json` versions.

Will be very handy for releases.


vinnyjth:
@redreceipt I've had a lot of luck with this approach in the past, I ask `npm` directly for the latest version of a package.

redreceipt:
yea it's just not super reliable to talk to NPM since we'll never know if the beta version for any given package got bumped

vinnyjth (approved):

PR #185, opened 2020-09-30
Renames Android Package

redreceipt:


apolloschurchapp/android/app/src/main/res/values/strings.xml
```diff
@@ -1,3 +1,3 @@
 <resources>
-    <string name="app_name">Apollos Church</string>
+    <string name="app_name">Apollos</string>
```
redreceipt, line 2:
```suggestion
    <string name="app_name">Apollos Church</string>
```

vinnyjth (approved):

PR #188, opened 2020-10-02
Simplifies Secrets Script

redreceipt:
This makes the bash script actually good code :)


scripts/secrets.sh
```diff
@@ -1,22 +1,31 @@
...
+
+SECRETS=(
+	"apolloschurchapp/.env.shared"
+	"apollos-church-api/.env.shared"
+	"apolloschurchapp/android/key.json"
+	"apolloschurchapp/android/app/apollos.keystore"
+)
+
+for file in "${SECRETS[@]}"; do
+	if [ "$1" = "-e" ]; then
```
christinamegura, line 23: Where would this "-e" come from?

christinamegura:
I'm new to looking at bash scripts but it is very easy to read and follow. How would I go about testing this?

redreceipt:
Awesome! Yea that -e is the first command line argument passed to the script. You would test this with yarn unpack-secrets <KEY> and then investigate the four files that were decrypted. The key comes from a file I put in Apollos Support project in Basecamp.

christinamegura:
`yarn unpack-secrets -e` encrypts all the files 👍 

`yarn unpack-secrets <KEY>` decrypts mostly everything except for about half of the apollos.keystore file. I don't know what to make of that

redreceipt:
Well technically you encrypt with yarn pack-secrets but maybe that’s confusing and could be better?

christinamegura:
Ope, duh! How do you feel about `yarn secrets`(...-e; ...<KEY>) instead? `pack` with `-e` might be a little redundant? 

Should the entire apollos.keystore file decrypt? Only a section of it did when I ran the command.

redreceipt:
you don't have to specify the -e. it's just 

`yarn pack-secrets <KEY>`
`yarn unpack-secrets <KEY>`

yea I'm not suprised a portion of the keystore is further encrypted, that's its purpose on Google's end

redreceipt:
maybe this clears it up, to answer your original question, that -e is built into the `yarn pack` command, see the package.json

redreceipt:
I made that change to the API, I think you're right, I like that better.

christinamegura:
Okay! 5th time is the charm.

`yarn unpack-secrets <KEY>` decrypts the four files, with apollos.keystore having a partial decrypt ✅

`yarn pack-secrets <KEY>` creates four new (edited) encrypted files ✅

This is clear and easy to follow. I'm chalking the confusion up to a personal problem since this is my first swing at a bash script and I didn't check everywhere I needed to at first.

christinamegura:
I like the API change but now when I open the four files they are already decrypted (not sure why this is happening now?). `yarn secrets -e <KEY>` will encrypt the files but then saves them as encoded (ie env.shared.enc). I don't think this was the case before and I cannot confirm the decrypt is working.


If you revert back to `pack-secrets` and `unpack-secrets` I have confirmed that working as follows:
- `yarn unpack-secrets <KEY>` decrypts the four files, with apollos.keystore having a partial decrypt 
- `yarn pack-secrets <KEY>` creates four new (edited) enctyped files

redreceipt:
What you described is how it’s supposed to work

christinamegura (approved):

PR #195, opened 2020-10-13
Simplifies Rock Import

redreceipt:


redreceipt:
@vinnyjth if we merge this one, we won't need the other "import" PR

vinnyjth:
I theoretically like this, but if you look at all of our client apps https://github.com/lcbc-digital/ephesus/blob/master/apollos-church-api/src/data/index.js#L16-L33 https://github.com/Differential/grace-apollos-app/blob/master/apollos-api/src/data/index.js#L17-L40

One of the first thing we do is comment out one or more of these modules and re-implement them. That will be easier if we leave all these explicit imports.

redreceipt:
As long as spread rock is listed first, they should be able to overwrite no problem right?

vinnyjth:
> As long as spread rock is listed first, they should be able to overwrite no problem right?

Totally. But it doesn't help when you want to comment out a data object with no replacement (I did this a few times w/ lcbc) and the list of data objects is a nice reference when remembering the naming for replacement.

vinnyjth:
I like the idea of some "catch all" to catch new data objects, but these days we tend to add them pretty rarely. Not having a spread by default also lets us add new (potentially unfinished) dataSources and not cause unexpected bugs / issues on client apps that pull canary builds for whatever reasons.

redreceipt:
Yea you're right

PR #196, opened 2020-10-13
home feature feed

vinnyjth:


apolloschurchapp/src/tabs/home/Home.js
```diff
@@ -29,43 +52,25 @@ class Home extends PureComponent {
...
+                    openUrl={openUrl}
+                    navigation={this.props.navigation}
+                    nodeId={data?.homeFeedFeatures?.id}
+                    onPressActionItem={handleOnPress}
+                    ListHeaderComponent={
+                      <LogoTitle source={require('./wordmark.png')} />
+                    }
+                  />
+                )}
+              </Query>
```
redreceipt, line 73: this makes me so happy

vinnyjth: 😂  me too

redreceipt:
looks good, blocked until the core PR makes it into canary

redreceipt (approved):

vinnyjth:
Looks like the canary build made it out.

vinnyjth:
@redreceipt What's the correct step now? Swap the package.json to point at the canary packages?

redreceipt:
yep, we can run `yarn canary` to update packages and test pointed to the live server

redreceipt:
not sure why the "check api" workflow isn't running, that should tell us it's ready

PR #201, opened 2020-10-16
Add the search feed back in

vinnyjth:


vinnyjth:
Future PR I'll move the SearchFeed / SearchInput into a connected component.

redreceipt (approved):

PR #203, opened 2020-10-16
bumps datetime picker version

redreceipt:
Screen Shot 2020-10-16 at 1 47 27 PM


vinnyjth:
Should we bump `react-native-modal-datetime-picker` to 9 as well? Technically I think we need to, but it seems to work fine without doing so.

redreceipt:
I mean it couldn’t hurt

vinnyjth (approved):

PR #209, opened 2020-10-28
Bumps to 2.0

redreceipt:


vinnyjth (approved):

redreceipt:
@vinnyjth wanna take a peek at this if you get a free minute? I've been trying to get these tests to pass for two days, it's why I killed the home tab test thinking that would help but no luck.

PR #210, opened 2020-10-28
fixes discover feed test

redreceipt:


apolloschurchapp/src/tabs/discover/__snapshots__/discover.tests.js.snap
```diff
@@ -437,14 +437,104 @@ exports[`The Discover tab component Should retrieve the Content Channel Feeds 1`
       </View>
     </View>
     <RCTScrollView
+      ListEmptyComponent={
+        <Err
+          error={
+            [Error: Network error: No more mocked responses for the query: query getFeatureFeed($featureFeedId: ID!) {
```
vinnyjth, line 443: Looks like this query is still erroring, but there's more data here than before.

redreceipt:
ok so now we're back to where we started

redreceipt:
any idea why we're getting a loading state here?

redreceipt:
@vinnyjth if you're cool with worrying about the load state later, the only reason I touched was to attempt to fix the strange jest error in the bump 2.0 PR.

vinnyjth:
We can fix it later. Fixing that issue will requiring writing some pretty intense mocks.

redreceipt:
can I get that approval

vinnyjth:
I wish when I expressed positive emotions about a PR Github would ahead and approve it for me automatically.

vinnyjth (approved):

redreceipt:
neuralink will make that happen soon enough

richarddubay:
> I wish when I expressed positive emotions about a PR Github would ahead and approve it for me automatically.

I feel like that's something Mergify should do! If I say "It's good" in a comment it just merges it. 😂

PR #211, opened 2020-10-28
Removes Home Feed Test

redreceipt:
Gives the home feed the "discover tab test" effect


redreceipt:
@vinnyjth only way I could get this to pass was to do the same for the home feed, delete the test

vinnyjth (approved):

PR #213, opened 2020-10-28
Fixes Tests on Next

redreceipt:
attempts to fix the tests


vinnyjth:
Getting this error locally (look like CI is as well) 

Screen Shot 2020-10-28 at 3 05 51 PM

redreceipt:
ok one sec, think I know what to do. thank God for the CI

redreceipt:
@vinnyjth ok, I just removed the discover feed test. it's the only one causing problems and it's invalid anyway so I figured we can cover it later if we so desire. it's one of our dependencies screwing it up I think anyway. probably because of conflicting RN versions or something.

PR #216, opened 2020-10-29
Switches from Docker to Standard Heroku

redreceipt:
I changed the default `yarn start` command in the API because Heroku can read that automatically with no config for Node.js servers


vinnyjth (approved):
I dig it. I like that `start` from root still runs `dev` but the `start` script in `api` runs `prod`

PR #220, opened 2020-11-06
fix: Don't error when formatting errors in dev mode.

vinnyjth:


redreceipt (approved):
haha nice.

PR #221, opened 2020-11-09
Adds safety catch to bump script

redreceipt:


scripts/bump.sh
```diff
@@ -19,10 +19,17 @@ VERSION=$(
...
+if [[ "$VERSION" == *beta* ]]; then
 	TAG=beta
+	PKG=$(npm show @apollosproject/config@beta version)
 else
 	TAG=latest
+	PKG=$(npm show @apollosproject/config version)
+fi
+
+if [[ "$VERSION" != "$PKG" ]]; then
+	echo "Latest Github tag ($VERSION) and NPM version ($PKG) doesn't match"
```
vinnyjth, line 31: great idea!

vinnyjth (approved):

PR #223, opened 2020-11-18
Simplifies CI Workflow

redreceipt:
Consolidates Actions workflows and uses the right convention. Now jobs are named correctly according to GH standards.


redreceipt:
@vinnyjth if this one looks good to you, I'll swap out the right "required" actions to pass

vinnyjth (approved):

PR #226, opened 2020-11-30
Adds postinstall at root

redreceipt:
This will run the postinstall scripts for each of the directories when running yarn from the root. Makes dev a little smoother


redreceipt:
@vinnyjth how you feel about this one? I think it'll make our lives easier

vinnyjth (approved):

PR #235, opened 2021-01-05
progress on update docs

vinnyjth:


redreceipt:
would it be better to make this a blog post? Easier to find and read for everyone I think

redreceipt:
@vinnyjth safe to assume we ain't gonna do this?

vinnyjth:
yep :)

PR #236, opened 2021-01-05
Adds Android Tab Bar Spacing

redreceipt:
Group 1


vinnyjth:
Do think we could move this component to the ui-kit?

vinnyjth (approved):
Looks good 👍  If we move this to ui-kit, we can do that in another PR.

PR #237, opened 2021-01-06
Fixes API Health Check

redreceipt:
This fixes the health check, it's been failing. It also adds in some debug to see what may be happening with the Gemfile on deploy


vinnyjth (approved):

redreceipt:
not quite right yet

PR #250, opened 2021-01-27
Cleans up Android URL

redreceipt:


vinnyjth (commented):
@redreceipt did you get a chance to test this? I'll need to pull it down, but from what I remember you do need this code in order to connect to the api properly on android sims / dev devices

redreceipt:
thought I did but it was a while ago, I'll double-check and re-tag you when I've tested.

redreceipt:
you are correct, not sure why I thought this would work

here's the article https://developer.android.com/studio/run/emulator-networking

PR #263, opened 2021-02-10
Create a map of types to interfaces and include in Config

jimmyhogoboom:

Creates a map of types to their interfaces and includes them in the ApollosConfig. Added to support https://github.com/ApollosProject/apollos-apps/pull/1728


vinnyjth:
@redreceipt This is a required step in upgrading to 2.0.1 btw

vinnyjth (approved):

redreceipt:
what do you mean "required" like a breaking change?

vinnyjth:
Yah, you will see errors in your app without it.

vinnyjth:
Which means that 2.0.1 should probably be 2.1.0

redreceipt:
is there a way we can make the apps change non-breaking before releasing 2.0.1? or else technically it belongs in v3 right?

PR #266, opened 2021-02-15
remove theme from API template

redreceipt:


redreceipt:
contingent on apps PR

PR #268, opened 2021-02-23
chore: Allow code to be manually deployed from specific branch

vinnyjth:


redreceipt:
this will most likely fail every time because our build numbers are driven by number of commits

vinnyjth:
@redreceipt I'm okay if the app build fails, I really want to be able to deploy to heroku.

vinnyjth:
I went ahead and split the deploys steps apart :)

redreceipt:
`git push heroku master` should do it?

I think the deploy settings need to stay set to “Heroku Git” for our deploy to work automatically

.github/workflows/deploy_client.yml
```diff
@@ -1,4 +1,4 @@
-name: Deploy
+name: Deploy Client
```
redreceipt, line 1:
```suggestion
name: Deploy Mobile App
```

vinnyjth:
@redreceipt I tried that. Doesn't work since our API code is in a sub folder. Our Github action accounts for that.

image

redreceipt (approved):

PR #277, opened 2021-03-03
feat: Follow suggestions

vinnyjth:


vinnyjth:
Yep! That sounds right! I think that might have been an issue with this branch not being up to date with master, since `Follow` is in master

PR #281, opened 2021-03-09
add search to home

redreceipt:
Screen Shot 2021-03-09 at 10 43 45 AM


apolloschurchapp/src/tabs/home/index.js
```diff
@@ -1,3 +1,16 @@
...
+
 import Home from './Home';
+import Search from './Search';
+
+const { Navigator, Screen } = createNativeStackNavigator();
+
+const HomeNavigator = () => (
+  <Navigator initialRouteName="Home">
+    <Screen component={Home} name="Home" options={{ headerShown: false }} />
+    <Screen component={Search} name="Search" options={{ headerShown: false }} />
```
vinnyjth, line 12: iirc the search screen needs to be moved to the root modal and this navigator is no longer needed
https://github.com/NewSpring/corinth/commit/942b67ec1655b6549088980d999352d8b4ae8401

(note, search is named "discover on NS")

vinnyjth (changes requested):
Thanks for doing this! One tweak I think we need, which I mentioned in a comment

apolloschurchapp/src/tabs/home/index.js
redreceipt: that was because the one in NS is a modal, this is just a stack native nav and having it here seems like it makes more sense to me because it's on the home screen. this is how I did LH.

vinnyjth: Ah, that makes more sense. @conrad-vanl do you care if the default for the search screen is a `modal` or a `stack`?

redreceipt:
@vinnyjth look now

vinnyjth (approved):

PR #282, opened 2021-03-09
add local onboarding images

redreceipt:
This will make setting up new apps faster as this is the recommended method for adding onboarding images

Group 1


redreceipt:
not sure the files changed tab is showing empty files for images...

vinnyjth (commented):
Good call, though I will miss those random splashes.

vinnyjth (approved):

redreceipt:
@vinnyjth I added another image and dismissals was still on so I'll need another approval here

vinnyjth (approved):

PR #290, opened 2021-03-16
simplify test mocks

redreceipt:
Better organizes mocks so they are easier to find


vinnyjth:
I dig it! Would love to go even further and move all of the `@apollosproject` ones to our ui-test-utils packag.e

vinnyjth (approved):

redreceipt:
> I dig it! Would love to go even further and move all of the `@apollosproject` ones to our ui-test-utils packag.e

I _think_ I like that 😂 not sure what you mean though

PR #300, opened 2021-03-18
remove testing panel

redreceipt:
We don't really use it and it's a big chunk of template code we could remove. The only thing in there is onboarding which can be launched by reinstalling the app


vinnyjth:
I'mma push back on this one :) it's way easier to test onboarding with this, and it all lives in it's own isolated folder.

redreceipt:
I figured you might :) you're the only one that uses it, I just figured stories get us what we need 90% of the time, and reinstall gets you onboarding. but I won't fight you, it is self-contained.

PR #301, opened 2021-03-18
switch search button to new useTheme hook

redreceipt:
Screen Shot 2021-03-18 at 4 50 54 PM

### DIFF
--- a/apolloschurchapp/src/Providers.js
+++ b/apolloschurchapp/src/Providers.js
@@ -30,6 +30,7 @@ const AppProviders = (props) => (
           <LiveProvider>
             <Providers
               themeInput={customTheme}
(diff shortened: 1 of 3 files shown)


apolloschurchapp/src/Providers.js
```diff
@@ -30,6 +30,7 @@ const AppProviders = (props) => (
           <LiveProvider>
             <Providers
               themeInput={customTheme}
+              customTheme={customTheme}
```
redreceipt, line 33:
```suggestion
              theme={customTheme}
```

redreceipt:
```suggestion
              theme={customTheme}
```

PR #303, opened 2021-03-19
feat: Set apollos user to true on login

vinnyjth:
This will tag users as apollos users when they open the app, login, and register. Not a great long term solution, but it should work very well for our needs right now.

### DIFF
--- a/apollos-church-api/src/data/rockWithPostgres.js
+++ b/apollos-church-api/src/data/rockWithPostgres.js
@@ -74,7 +74,6 @@ export const resolver = {
             value: true,
           },
         ]);
-        console.log('setting apollos user');
       } catch (e) {
         console.warn(e);
       }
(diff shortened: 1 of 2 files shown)


apollos-church-api/src/data/rockWithPostgres.js
```diff
@@ -56,5 +56,31 @@ export const resolver = {
...
+      }
+
+      try {
+        await dataSources.Person.updateProfile([
+          {
+            field: 'apollosUser',
+            value: true,
+          },
+        ]);
+        console.log('setting apollos user');
```
vinnyjth, line 77:
```suggestion
```

apollos-church-api/src/data/rockWithPostgres.js
```diff
@@ -56,5 +56,31 @@ export const resolver = {
         { field: 'campusId', value: campus.id },
       ]); // updates in Postgres
     },
+    updateUserPushSettings: async (root, { input }, { dataSources }) => {
```
vinnyjth, line 59: This function is copied over from `oneSignalWithRock`

vinnyjth: It fires on app open and on `login`

jimmyhogoboom (approved):

PR #309, opened 2021-03-24
use apollo built-in health check

redreceipt:


apollos-church-api/src/server.js
```diff
@@ -74,22 +71,6 @@ const apolloServer = new ApolloServer({
 
 const app = express();
 
-// health check
-app.get('/health', cors(), (req, res) => {
-  res.send('ok');
-});
-
-// apollos version
-app.get('/version', cors(), (req, res) => {
```
vinnyjth, line 83: Do you remember why we had this version check in here?

vinnyjth (approved):

redreceipt:
Yea I put I there when we were gonna do an admin panel with a status check of all the apps. But we can just look at the Apollos.JSON file if we ever get around to doing something like that.

redreceipt:
@vinnyjth I just removed sqlite completely in this PR, that's ok right?

vinnyjth:
> @vinnyjth I just removed sqlite completely in this PR, that's ok right?

Yep! that should be good, we've moved past supporting it as a default.

PR #310, opened 2021-03-24
Add package only adds updated packages

vinnyjth:
My shell foo is bad so I rewrote the script in node. 

The script goes and compares the current package.json version with the version of the tag. It only updates the packages if the tag version is greater. 

This fixes an issue where I was going in and having to manually update most packages to the latest stable version every time I ran `yarn canary`, since most canary packages aren't updated every release (they would get bumped to an older release when running yarn canary) 

The script might take a wee bit longer if all the packages need updating, but in my testing it actually takes way less time since most of the time only a few packages need to be updated at once.

### DIFF
--- a/scripts/add-packages.js
+++ b/scripts/add-packages.js
@@ -1,0 +1,90 @@
+const fs = require('fs');
+const { execSync, exec } = require('child_process');
+
+const semver = require(`semver`);
+
+const tag = process.argv[2];
+const force = process.argv[3];
+
+if (!tag) {
+  console.warn('Usage: ./add-packages.sh <TAG>');
+}
+
+const packageJson = JSON.parse(fs.readFileSync(`${process.env.PWD}/package.json`));
+
+const dependencies = [...Object.keys(packageJson.dependencies)]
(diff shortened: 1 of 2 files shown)


redreceipt (approved):

redreceipt:
actually all we'd have to do is force publish another canary before a stable release so 2.5.0-canary.15 would match 2.5.1

vinnyjth:
> actually all we'd have to do is force publish another canary before a stable release so 2.5.0-canary.15 would match 2.5.1

I would rather do that :) 

However, something to consider is that this approach is a bit quicker to run (not a big deal) and makes it clear which packages are actually on real canary versions (kinda nice)

redreceipt:
which actually makes more sense i think because then you'd know by pulling canary, you were getting the absolute latest every time

redreceipt:
i'm gonna go do that regardless of what we do here actually, better for clients to be able to say add @canary and always get latest

redreceipt:
making draft, reopen pending the apps canary version thing doesn't work right

PR #313, opened 2021-03-24
add basic e2e testing

redreceipt:


redreceipt:
right now we can't build a release to a sim. probably would work with a debug build but that kinda defeats the purpose. pick back up once we can succesfully run `react-native run-ios --configuration Release`

redreceipt:
release builds breaking due to one signal, maybe upgrading to v4 will fix...

PR #314, opened 2021-03-24
remove Google Cast

redreceipt:
Removes all the infrastructure required for Google cast, there was still surprisingly a lot. This gives us back a couple of permissions and maybe will other things smoother. Fingers crossed, this fixes being able to put release builds on a sim, the error currently is with Google Cast.


redreceipt:
@vinnyjth @conrad-vanl the only one I'm unsure about, I removed the Wifi capability but does Airplay need that?

apolloschurchapp/ios/apolloschurchapp/apolloschurchapp.entitlements
```diff
@@ -4,8 +4,6 @@
 <dict>
 	<key>aps-environment</key>
 	<string>development</string>
-	<key>com.apple.developer.networking.wifi-info</key>
```
redreceipt, line 7: may still need this for Airplay, I'm not sure...

vinnyjth: I think this is just a cast thing

vinnyjth (approved):

PR #319, opened 2021-03-29
remove NodeSingle

redreceipt:
The `OPEN_NODE` action has been deprecated which was the only reference to `NodeSingle`. And it's pretty much a less functional `ContentSingle`


redreceipt:
blocked version 2.7.0+

vinnyjth (approved):

PR #327, opened 2021-04-14
use shared config instead of production file

redreceipt:
Realized my `sed` script stopped working, not sure why, but it made me realize we don't really need it anymore


redreceipt:
@vinnyjth can you take a look at this one, it's fairly critical as we can't log in to the core app and may explain why the LH Android app can't login either

vinnyjth (approved):

PR #331, opened 2021-04-27
simplified bump script

redreceipt:


vinnyjth (approved):

vinnyjth (changes requested):
image

vinnyjth:
Looks like the output is slightly different. We might need to strip that "v" off the front

redreceipt:
shoot you're right

scripts/bump.sh
```diff
@@ -7,14 +7,8 @@ curl 'https://api.github.com/repos/apollosproject/apollos-apps/tags'
...
-	curl -s 'https://api.github.com/repos/apollosproject/apollos-apps/tags' |
-		python -m json.tool |
-		grep '\"name\"' |
-		cut -d ':' -f 2 |
-		sed 's/&quot;/\"/g' |
-		sed -E 's/\"v(.*)\".*/\1/g' |
-		sed -n 1p |
-		tr -d '[:space:]'
+	curl -s "https://api.github.com/repos/apollosproject/apollos-apps/tags" |
+		python -c "import sys, json; print json.load(sys.stdin)[0]['name']"
```
redreceipt, line 11:
```suggestion
		python -c "import sys, json; print json.load(sys.stdin)[0]['name'][1:]"
```

redreceipt:
actually now I'm worried about the carriage return 😂 just a sec

redreceipt:
ok i think this will work

vinnyjth:
@redreceipt yah I ran some tests locally and things looked fine - the `npm` command also has a carriage return.

vinnyjth (approved):

PR #338, opened 2021-05-03
initialization improvements

redreceipt:
This attempts to make the setup process smoother in various ways


redreceipt:
this will be the only big one, I'll do the rest in small chunks

vinnyjth (approved):
Excellent! Very pumped to have more of this

PR #341, opened 2021-05-04
remove dummy suggested follows

redreceipt:


vinnyjth (changes requested):
any chance you could leave these in but comment them out? or use fake emails? since this option is handy but not documented anywhere else

redreceipt:
> any chance you could leave these in but comment them out? or use fake emails? since this option is handy but not documented anywhere else

what if I write a simple markdown doc and drop it in?

vinnyjth:
> > any chance you could leave these in but comment them out? or use fake emails? since this option is handy but not documented anywhere else
> 
> what if I write a simple markdown doc and drop it in?

Sounds fantastic. Might be nice if we could have a `config.yml.md` that documented all of the important/confusing variables

redreceipt:
yes that's a good idea. I'll start one where we can drop documentation for any of the variables that need it

redreceipt:
ok @vinnyjth take a look

vinnyjth:
stellar !

vinnyjth (approved):

PR #345, opened 2021-05-04
remove unneeded default config variables

redreceipt:


apollos-church-api/config.yml
```diff
@@ -24,48 +23,26 @@ DATABASE:
         require: false
         rejectUnauthorized: false
 BUGSNAG:
-  ## Optional, but you'll need to remove references to @apollosproject/bugsnag if you don't wish to use.
   API_KEY: ${BUGSNAG_KEY}
 ROCK:
   URL: ${ROCK_URL}
   API_TOKEN: ${ROCK_TOKEN}
-  # This should match the timezone of the Rock server
-  TIMEZONE: 'America/New_York'
```
vinnyjth, line 33: I think this one is still required 🤔 

https://github.com/ApollosProject/apollos-apps/search?q=ROCK.TIMEZONE

apollos-church-api/config.yml
```diff
@@ -24,48 +23,26 @@ DATABASE:
...
 UNIVERSAL_LINKS:
-  # This should be your app's bundle ID, as found in your info.plist.
```
vinnyjth, line 35: Why remove these comments?

apollos-church-api/config.yml
```diff
@@ -24,48 +23,26 @@ DATABASE:
...
   APP_ID: ${ONE_SIGNAL_APP_ID}
   REST_KEY: ${ONE_SIGNAL_REST_KEY}
 CHURCH_ONLINE:
   URL: ${CHURCH_ONLINE_URL}
-  MEDIA_URLS: []
   WEB_VIEW_URL: https://apollos.online.church/
 ALGOLIA:
   APPLICATION_ID: ${ALGOLIASEARCH_APPLICATION_ID}
   API_KEY: ${ALGOLIASEARCH_API_KEY}
-  SEARCH_INDEX: prod_ContentItem
```
vinnyjth, line 56: 👍

apollos-church-api/config.yml
```diff
@@ -78,25 +55,13 @@ PASS:
...
-  SEGMENT_KEY:
-  GA_ID:
-  USE_ROCK: false
 BIBLE_API:
   KEY: ${BIBLE_API_KEY}
   BIBLE_ID:
     WEB: '9879dbb7cfe39e4d-01'
     KJV: 'de4e12af7f28f599-02'
 CLOUDINARY:
-  # Cloudinary is used to improve image loading
```
vinnyjth, line 93: I think this is another comment that could stay :) The key is pretty ambiguous if you aren't familiar with cloudinary

apollos-church-api/src/config.js
```diff
@@ -7,6 +7,11 @@ ApollosConfig.loadYaml({
   configPath: path.join(__dirname, '..', 'config.yml'),
 });
 
+// defaults
+ApollosConfig.loadJs({
```
vinnyjth, line 11: oh doh 😂 disregard my comment above

vinnyjth (approved):

apollos-church-api/src/config.js
redreceipt: honestly, this was an experiment, we could throw more in here if they make sense. the goal I think for me is to be completely driven off ENV and not need config 😬

apollos-church-api/config.yml
redreceipt: I removed this comment (and others) because I describe what it is and how to get it (ENV variable) in the cloudinary set up doc

redreceipt: so you never actually look at this comment anyway

apollos-church-api/src/config.js
vinnyjth: totally. Maybe we need to formalize this a bit more? config.yml for values that NEED configuration, .env for values that vary per environment or are secret, and blocks like this for variables w/ sane defaults?

apollos-church-api/config.yml
redreceipt: same answer as below

PR #351, opened 2021-05-10
Add one signal notification service extension

vinnyjth:


vinnyjth:
https://user-images.githubusercontent.com/1637694/118181602-ef721780-b405-11eb-8893-efebadf69cc1.MP4

apolloschurchapp/fastlane/Fastfile
```diff
@@ -13,12 +14,21 @@
 # Uncomment the line if you want fastlane to automatically update itself
 #update_fastlane
 
+def product_bundle_id(path:)
+  project = Xcodeproj::Project.open(path)
+  scheme = project.native_targets.first
+  build_configuration = scheme.build_configurations.first
+  build_configuration.build_settings['PRODUCT_BUNDLE_IDENTIFIER']
```
vinnyjth, line 21:
```suggestion
  build_configuration.build_settings['PRODUCT_BUNDLE_IDENTIFIER'].strip
```

vinnyjth:
@redreceipt We'll need to merge this as well

apolloschurchapp/fastlane/Fastfile
```diff
@@ -27,9 +36,11 @@ platform :ios do
     # get project name
     proj = Dir.glob("../ios/*.xcworkspace")[0][7..-13]
 
-    app_store_connect_api_key(key_filepath: "ios/apollos.p8")
+    identifier = product_bundle_id(path: "../ios/" << proj << ".xcodeproj")
+
+    app_store_connect_api_key(key_filepath: "ios/apollos.p8", in_house: false)
     setup_ci
-    match(type: 'appstore', readonly: true)
+    match(type: 'appstore', app_identifier: [identifier, "#{identifier}.OneSignalNotificationServiceExtension"])
```
vinnyjth, line 43:
```suggestion
    match(type: 'appstore', app_identifier: [identifier, "#{identifier}.OneSignalNotificationServiceExtension"], readonly: true)
```

vinnyjth: add readonly: true back

redreceipt:
shouldn't merge this until we have an apps release, I was trying to get your other one in before running a release

vinnyjth:
Oh yah, you're right.

apolloschurchapp/fastlane/Fastfile
```diff
@@ -27,9 +36,11 @@ platform :ios do
     # get project name
     proj = Dir.glob("../ios/*.xcworkspace")[0][7..-13]
 
-    app_store_connect_api_key(key_filepath: "ios/apollos.p8")
+    identifier = product_bundle_id(path: "../ios/" << proj << ".xcodeproj")
+
+    app_store_connect_api_key(key_filepath: "ios/apollos.p8", in_house: false)
```
redreceipt, line 41: what's this `in_house` for?

vinnyjth:
It tells iTC we don’t want to distribute in house (enterprise). I ran into an error testing distributing that this flag fixes.

redreceipt (dismissed):

redreceipt:
@vinnyjth before we merge this, I remember now why I made the bundle ID come from env, when you're generating certs using the command line, it comes from the env. Plus if we make it implicit like this, new church apps freshly cloned are pointed at the core template app bundle ID which is wrong.

how do you feel about `MATCH_BUNDLE_ID=com.differential.apollos,com.differential.apollos.OneSignal` like it was before

vinnyjth:
How does `MATCH_BUNDLE_ID` end up being used in the `Fastfile`? I'm fine with that, if we can do what you're describing, but I'm struggling to find any documentation on `MATCH_BUNDLE_ID`. 

beyond that I like the explicitness of having the bundle ids called out in the `Fastfile`, makes it clear what's going on 

Shouldn't churches get new bundle ids almost immediately as part of `yarn setup`?

redreceipt:
it's actually `MATCH_APP_IDENTIFIER` and that's what we currently use. It is a part of the init script, that variable is set.

redreceipt:
I moved everything out of the `Fastfile` into env so that it doesn't reflect anything from core. And I like that it would fail if they haven't set the variable, there's no way to accidentally regen certs for core (something Nathan would have done if he had rights to the Di app)

redreceipt:
all of these are currently required for our fastlane setup

Screen Shot 2021-06-17 at 11 55 55 AM

vinnyjth:
I see that - but the build would fail anyway if they had the wrong bundleId set in the info.plist right? You can't have the BUNDLE_ID mismatch the Bundle ID in the Info.plist, and so the code I have would remove our meed for MATCH_APP_IDENTIFIER and wouldn't have additional risk for pulling in config from core.

vinnyjth:
It essentially mimics what Android already does (which is a different ballgame, certainly)

redreceipt:
yea I see your point, I like removing the redundancy. But when going through it, new devs may not have changed the bundle ID in Xcode yet. That's what I'm worried about.

redreceipt:
I suppose, worst case, a Di dev overwrites the core app certs. We can cross that bridge when we come to it. I like the change.

redreceipt (approved):

vinnyjth:
fair - so looking at this from two perspectives
- What's the likelyhood that they will have an invalid bundleId?: `react-native-rename` sets a new bundleId so if they follow our process, it's low 
- What if they reuse our existing bundle ID? Depends on what else they have setup. Either they will re-use our bundle ID (which would leave garbage in iTC, easy to delete though), pull down the existing certs, or generate new certs while deleting the existing ones in ITC. The last case would be the worst, but they would have to enter the Apollos Team data in their `.env` and we could easily fix. So I think it's pretty safe :)

PR #352, opened 2021-05-10
consolidates tabs into one component

redreceipt:


redreceipt:
@conrad-vanl this is the one

redreceipt:
alright @conrad-vanl ready to take a peek at

vinnyjth:
I can confirm that this only occurs on this branch, does not occur in master. The "next" button doesn't navigate, but it does send a text message / check if a user exists.

vinnyjth:
Fixed it :D Carry on

apolloschurchapp/src/tabs/Tab.js
```diff
@@ -0,0 +1,69 @@
...
+        }
+      }
+    `,
+    { variables: { tab }, fetchPolicy: 'cache-and-network' }
+  );
+
+  // this is only used by the tab loaded first
+  // if there is a new version of the onboarding flow,
+  // we'll navigate there first to show new screens
+  useEffect(
```
conrad-vanl, line 38: This seems like it should be rendered by the tab navigator, instead of each tab?

apolloschurchapp/src/tabs/index.js
```diff
@@ -1,27 +1,103 @@
...
+    screenOptions={{
+      headerHideShadow: true,
+      headerLargeTitle: true,
+    }}
+  >
+    <WatchStack.Screen name="Watch" component={Watch} />
+  </WatchStack.Navigator>
+);
+
+const PrayStack = createNativeStackNavigator();
```
conrad-vanl, line 65: Would like @vinnyjth thoughts on structuring all of these inside of one file instead of separated. I don't have strong opinion

conrad-vanl (commented):
image

Super quick code glance from my morning spot.

apolloschurchapp/src/tabs/Tab.js
redreceipt: Ooo yea I like that

apolloschurchapp/src/tabs/Tab.js
```diff
@@ -0,0 +1,87 @@
...
+
+import { ONBOARDING_VERSION } from '../ui/Onboarding';
+
+function handleOnPress({ action, ...props }) {
+  if (FEATURE_FEED_ACTION_MAP[action]) {
+    FEATURE_FEED_ACTION_MAP[action]({ action, ...props });
+  }
+}
+
+export const createTab = ({ tabName, screenOptions, ...props }) => {
```
conrad-vanl, line 23: Need to think really hard about what api and props this exposes. Right now I would say this feels inflexible. The Tab Component is really just a FeaturesFeed component, but if I didn't want a featuresfeed, I don't think I could change that?

apolloschurchapp/src/tabs/Tab.js
```diff
@@ -0,0 +1,87 @@
...
+  const TabStack = createNativeStackNavigator();
+  const TabComponent = () => <Tab tab={tabName.toUpperCase()} />;
```
conrad-vanl, line 25: Could this be an argument to the component somehow?

apolloschurchapp/src/tabs/Tab.js
```diff
@@ -0,0 +1,87 @@
...
+  const TabNav = () => (
+    <TabStack.Navigator
```
conrad-vanl, line 27: Are there other props one might want to pass to this navigator? Like additional routes...or other props

apolloschurchapp/src/tabs/Tab.js
```diff
@@ -0,0 +1,87 @@
...
+      screenOptions={{
+        headerHideShadow: true,
+        headerLargeTitle: true,
+        ...screenOptions,
+      }}
+    >
+      <TabStack.Screen name={tabName} component={TabComponent} {...props} />
```
conrad-vanl, line 34: How might I pass additional routes?

conrad-vanl: And what does spreading props here do differently then spreading props in the component def above?

redreceipt:
@vinnyjth since there seems to be a lot of questions around the edit (I posted my thoughts on slack) could we push what I had before and then work to make it even better in another PR?

apolloschurchapp/src/tabs/Tab.js
vinnyjth: I think it's fine if it's inflexible, I just need to rename to `createFeatureFeedTab`

apolloschurchapp/src/tabs/Tab.js
vinnyjth: yah, made an explicit prop

apolloschurchapp/src/tabs/Tab.js
vinnyjth: There might be, but for most cases no. Goal is to make this a focused factory-ish function and let the user go their own way if needed.

apolloschurchapp/src/tabs/Tab.js
vinnyjth: Removed spreading of props. If a user wants to pass additional props they shouldn't use this function.

vinnyjth:
I found a bug I need to fix before merging ...

vinnyjth:
simulator_screenshot_281BDE8A-F993-44D6-B3F3-4F857BFB1BCB

redreceipt:
> I found a bug I need to fix before merging ...

yea looks like we need a `useNavigation` somewhere

vinnyjth:
Fixed. Checking to see if the content channel screen also looks like this in `master`

simulator_screenshot_3C1EAC85-F2D6-4FEB-8170-FB7C47E24583

apolloschurchapp/src/tabs/Tab.js
```diff
@@ -73,6 +75,7 @@ const Tab = ({ tab }) => {
             openUrl={openUrl}
             featureFeedId={data?.tabFeedFeatures?.id}
             onPressActionItem={handleOnPress}
+            navigation={navigation}
```
redreceipt, line 78: instead of passing through, shouldn't this be on the component that needs it?

vinnyjth:
Nah, looks way better in `master` :/ 

image

apolloschurchapp/src/tabs/Tab.js
redreceipt: feels like you should never use props for something there's already context for

redreceipt:
> Nah, looks way better in `master` :/

needs a header:true somewhere

vinnyjth:
> > Nah, looks way better in `master` :/
> 
> needs a header:true somewhere

Looking at `master`, I think we need `ContentFeed` to be a child in each of the tab routes.

redreceipt:
> > > Nah, looks way better in `master` :/
> > 
> > 
> > needs a header:true somewhere
> 
> Looking at `master`, I think we need `ContentFeed` to be a child in each of the tab routes.

only if we want to see the tab bar

vinnyjth:
Given the slack convo, I think that's desirable? 

> A lot of apps have sub routes on tabs, I’d say they are more common in most apps then routes that navigate on top of the tabs

redreceipt:
@vinnyjth feel good about this?

vinnyjth (approved):

PR #366, opened 2021-05-19
adds server caching

redreceipt:
Does a couple of things:
1. Turns on server caching
2. Caches private scoped requests by using the auth token
3. Enables Automatic Persisted Queries for `GET` requests


redreceipt:
also when this goes into affect, we'll probably want to check for an existing Redis connection from the Redis DS before creating a new one

camrun91:
@redreceipt any idea when this will get pulled in? I was just about to do some of this on our repo but if it is close to being pulled in I may wait.

redreceipt:
I don't have a good answer for you :/ Certainly has value but is a low priority for us so not sure when the team will reach a conclusion on it.

camrun91:
OK cool I may try and implement on our side then as I was just about to do a lot of the same logic you did with this.

conrad-vanl:
@vinnyjth @redreceipt should we look back into this with LCBC stuff? I don't like the idea of dropping Fastly.... honestly I'd rather move to like [graphcdn](https://graphcdn.io) but in the meantime keeping queries conferred for GET requests seem like a sensible change in this PR to keep this non-breaking with current apps.

redreceipt:
gonna defer to @vinnyjth here, I don't know the advantage of this PR if we were to keep GET requests. Are we assuming POST is not capable of being used in production at scale? I thought that was the default for GraphQL?

conrad-vanl:
@redreceipt the changes in this PR are good to keep regardless if the requests are sent over GET or POST, but the problem is Fastly can't cache POST requests (GraphCDN can), and we don't know what the impact will be from dropping Fastly. We know that Fastly handled 90% of the traffic on Sundays for LCBC.... if your goal is to get rid of Fastly for all apps - HMW gain confidence that we can handle traffic at scale better than a CDN?

I'm currently of the believe that:

1. This is still better then not having it, since we gain some caching on per-user queries. Those are still expensive.
2. This, along with the Postgres work, means we won't need to use the more complex Redis key caching that we use now
3. We will still likely need to be able to leverage a CDN for larger applications, and unless we move to GraphCDN, that means maintaining GET support

camrun91:
@conrad-vanl I have this pr in our app that is currently in test flight if you are thinking a move to GraphCDN would be beneficial in addition to this then I can look at using it in ours and try and give some feedback in to this?

apolloschurchapp/ios/Podfile.lock
```diff
@@ -613,7 +618,7 @@ SPEC CHECKSUMS:
   boost-for-react-native: 39c7adb57c4e60d6c5479dd8623128eb5b3f0f2c
   BVLinearGradient: e3aad03778a456d77928f594a649e96995f1c872
   CocoaAsyncSocket: 065fd1e645c7abab64f7a6a2007a48038fdc6a99
-  DoubleConversion: cde416483dac037923206447da6e1454df403714
+  DoubleConversion: cf9b38bf0b2d048436d9a82ad2abe1404f11e7de
```
redreceipt, line 621:
```suggestion
  DoubleConversion: cde416483dac037923206447da6e1454df403714
```

apolloschurchapp/ios/Podfile.lock
```diff
@@ -624,7 +629,7 @@ SPEC CHECKSUMS:
   Flipper-RSocket: 127954abe8b162fcaf68d2134d34dc2bd7076154
   FlipperKit: 8a20b5c5fcf9436cac58551dc049867247f64b00
   Folly: b73c3869541e86821df3c387eb0af5f65addfab4
-  glog: 40a13f7840415b9a77023fbcae0f1e6f43192af3
+  glog: 73c2498ac6884b13ede40eda8228cb1eee9d9d62
```
redreceipt, line 632:
```suggestion
  glog: 40a13f7840415b9a77023fbcae0f1e6f43192af3
```

redreceipt:
Ok, @vinnyjth @conrad-vanl this one is ready again. I turned back on `GET` requests. And I also configured automatically persisted queries to cut down on the bandwidth coming from the client and ensuring we don't fail because the query strings are too long.

camrun91:
I can look at these this evening or tomorrow to pull the changes in our app. Will I need to add castle back or should this still work without?

redreceipt:
> I can look at these this evening or tomorrow to pull the changes in our app. Will I need to add castle back or should this still work without?

castle?

camrun91:
Dang auto correct that should have been fastly

redreceipt:
Oh yes, if you want to take advantage of Fastly caching the hashed GET queries, you'll need to turn it back on.

camrun91:
@redreceipt @conrad-vanl I have our main branch and test flight app using this now with Fastly. I do not know if there is any data or metrics or testing that would be of use to you all but if so I would be more than glad to help.

redreceipt:
@vinnyjth any reason we wouldn't consider this? it gets us POST caching and still uses GET for everything else.

redreceipt:
@vinnyjth which part of this are you still out on? there's a couple of churches using cached POST requests in the wild. And APQ simply hashes GET requests but nothing else changes.

redreceipt:
ping

redreceipt:
@vinnyjth to clarify in case there are still questions, this doesn't remove fastly

vinnyjth (approved):

PR #372, opened 2021-05-25
adds and centers new dark wordmark

redreceipt:


apolloschurchapp/src/tabs/index.js
```diff
@@ -11,9 +11,16 @@ import { ONBOARDING_VERSION } from '../ui/Onboarding';
 import Connect from './connect';
 import tabBarIcon from './tabBarIcon';
 
-const HeaderLogo = styled(({ theme }) => ({
-  height: theme.sizing.baseUnit,
-  resizeMode: 'contain',
+const HeaderLogo = withTheme(({ theme }) => ({
+  style: {
+    height: theme.sizing.baseUnit * 2.5,
+    width: '70%',
```
vinnyjth, line 17: do you need `width` here? It seems like a smell. Looking great beyond that :D

redreceipt: That's actually what centers the wordmark. Not sure why. Definitely a smell. Something to do with how react-navigation positions its header center item. Open to suggestions though.

vinnyjth: I imagine it's because the right hand icon takes up ~15% of the space, when the width is auto, the view takes up the whole space and the center ends up offset to the left. If you reduce the width, the center eeks over to the right until the width is smaller than (100% - search icon width*2)

vinnyjth: I'm fine merging this for now :) I'm trying to figure out why LCBC doesn't need this

vinnyjth (approved):

redreceipt:
You would think that, but even if my testing, omitting the right icon still leaves the icon off-center if I remember correctly

PR #373, opened 2021-05-25
adds simple load test script

redreceipt:
Screen Shot 2021-05-25 at 3 28 53 PM


redreceipt:
@vinnyjth another plus here is this will allow us to run integration tests on the API as part of the CI checks

PR #379, opened 2021-06-04
remove publish from workflow

redreceipt:
publish schema is done through env variable now, no command necessary


vinnyjth:
I was about to create a PR removing `get-schema` 👍

vinnyjth (approved):

PR #382, opened 2021-06-07
remove redundant typemap declaration

redreceipt:


vinnyjth:
I would merge this without a glance - but we are using the fact that these are separate bits on NS to have fine control over loading states - https://github.com/NewSpring/corinth/blob/master/loadConfig.js#L26

We need to find a better solution for what NS is forced to do before we can merge these.

redreceipt:
They could do the same thing, just spread in the type map from core and then overwrite those two, right?

vinnyjth:
@redreceipt That would break the `cache.js` file. We need the `cache.js` to be `fragmentTypes.json `and `ApollosConfig.TYPE_MAP` to be just override.

vinnyjth:
The better path might be to fix the loading state guesser so we don't need that override to begin with. Not sure we can though with our current schema.

redreceipt:
Oh shoot

redreceipt:
blocked on release

vinnyjth (approved):

PR #385, opened 2021-06-09
Testing Panel - Following List

tomthornton:


redreceipt:
@tomthornton you still need this follow list in the testing panel?

tomthornton:
@redreceipt not sure. are we planning on implementing this?

redreceipt:
If you don't care either, I'd opt to leave it out. Figured you had a reason for submitting.

tomthornton:
@redreceipt this was one of my first PRs. I think we should revisit it later and discuss some design changes

redreceipt:
yep I agree, sounds good

PR #389, opened 2021-06-12
move testing panel to user settings

redreceipt:
also deletes a lot of code and hides testing panel behind development environment so we don't have to worry about changing anything for new app spin ups

Screen Shot 2021-06-11 at 10 50 03 PM


redreceipt:
@vinnyjth what do you think about this one? it's "blocking" us from trying to remove `NavService` completely. it compresses the current test panel code into 50 lines and makes it turned off in prod by default so no more commenting out or deleting on client apps.

vinnyjth (approved):

PR #396, opened 2021-06-17
remove secrets script

redreceipt:


vinnyjth:
do you think we should move the secrets script into the individual scripts for `apollos-church-api` and `apolloschurchapp`?

redreceipt:
> do you think we should move the secrets script into the individual scripts for `apollos-church-api` and `apolloschurchapp`?

no need now with the CLI

redreceipt:
> That's what I meant rather, so the `yarn secrets` would be an alias for `npx @apollosproject/apollos-cli secrets`

Oh. I would rather not simply because it's cleaner and it's not used that often. Plus what if they have the CLI installed globally instead.

vinnyjth (approved):

PR #397, opened 2021-06-17
Upgrades to React Native 64

redreceipt:
Upgrade is minor, this puts us waaaay closer to the core react native project without a lot of unnecessary technical debt we've accrued.


redreceipt:
pending https://github.com/facebook/react-native/pull/31740

redreceipt:
@vinnyjth i got carried away last night 😬 it's actually A LOT closer to working than I thought on the first pass

apolloschurchapp/ios/Podfile
```diff
@@ -20,31 +24,21 @@ target 'apolloschurchapp' do
   # Enables Flipper.
   #
   # Note that if you have use_frameworks! enabled, Flipper will not work and
-  # you should disable these next few lines.
+  # you should disable the next line.
   use_flipper!('Flipper' => '0.75.1', 'Flipper-Folly' => '2.5.3', 'Flipper-RSocket' => '1.3.1')
```
vinnyjth, line 28: I think we can omit the "custom" flipper versions now, also. that was a patch for our previous version of RN

redreceipt: it's still listed in a thread I found somewhere as a problem with 64 too

redreceipt:
@vinnyjth this one is ready to look at. the bulk of the changes are linter changes. I took out our custom config and just went with react native community config now that they have one, removes _a lot_ of infrastructureure

vinnyjth:
@redreceipt I LOVE that we are able to eliminate our custom config, but.... this will wreak havoc with the update tool 😬 . It will likely force manual merges for any files that have been modified by clients. Obviously, that's okay! It's what the update tool is for, but does the costbenefitt work out in our favor given that the custom linter config rarely causes issues? 

As integration lead, you get to make that call :) I like the idea of waiting until we've got 98% of the code in `apps` before touching our lint settings.

redreceipt:
Oh, shoot I didn't think about that. Honestly, I did it because it puts us way closer to a true RN project which makes future `react-native-upgrade` calls easier. And if I push it off, I'll have to do the hard work again later of manually figuring out which of our eslint deps and config files are necessary and which are custom. I think I'd rather just bite the bullet now. App directories are already barely custom save the older apps. I think it'll be obvious to folks that `REMOTE` is just linter changes. We should just be able to always choose `HEAD` and then run `eslint --fix`, right?

vinnyjth:
> I think it'll be obvious to folks that REMOTE is just linter changes. We should just be able to always choose HEAD and then run eslint --fix, right?

Ooohh that's a good point. That's assuming they are just moving one version up. If they are updating a bunch of different versions, with multiple template changes, they'll have to keep track of what files need to go to HEAD and what files need to be actually merged. So maybe we tell people to update to 2.29.1 and then from there to 2.30? 

What if this is the start of version 3?

redreceipt:
I don’t think it should be v3, that seems like it should be for NPM breaking changes. I like upgrading to 2.29 and then 2.30. almost all of our apps are caught up anyway and the ones that aren’t we’ll know to do that.

redreceipt:
ok @vinnyjth I got rid of most of the linter differences. Maybe we keep curly braces around if statements though?

Screen Shot 2021-08-12 at 7 42 14 PM

redreceipt:
blocked by https://github.com/ApollosProject/apollos-apps/pull/2063 and https://github.com/ApollosProject/apollos-templates/pull/430 so the events tests and mocking the timezone will be a non-issue

PR #409, opened 2021-06-23
Update references to `yarn secrets`

rkdavidson:
# About
PR #396 removed the `yarn secrets` script in favor of `npx @apollosproject/apollos-cli secrets`.
This PR updates references in the readme and github actions config.

_EDIT_ Looks_ like @nlewis84 beat me to the readme changes in #408 ! 

# Tradeoffs/Caveats
⚠️  I am not sure what the impact is on github actions, here. Locally, the first time I ran the `npx` script, I was prompted to install the Apollos CLI tools. If that step is already being taken care of (or a prerequisite) of the github CI pipeline, it's _probably_ safe. Otherwise, we may need to ensure Apollos CLI is installed prior to invoking the script?

Looking for guidance on the team, here ☝️ .

Sorry for the whitespace/auto formatting changes. I can revert those, if need be. Didn't notice my IDE did that until opening this PR.


.github/workflows/deploy.yml
```diff
@@ -52,35 +52,35 @@ jobs:
...
 
-    - name: Install Bundler
-      run: gem install bundler
+      - name: Install Bundler
+        run: gem install bundler
 
-    - name: Decrypt Environment Variables and Apple API Key
-      run: (cd .. && yarn secrets -d ${{ secrets.ENCRYPTION_PASSWORD }})
+      - name: Decrypt Environment Variables and Apple API Key
+        run: (cd .. && npx @apollosproject/apollos-cli secrets -d ${{ secrets.ENCRYPTION_PASSWORD }})
```
vinnyjth, line 66: Small tweak here :) I believe you can remove the `cd ..` since the script needs to be run alongside the .env files.

rkdavidson: Good catch! I can make that change, but the `cd ..` was there before — we're confident that's safe to update?

vinnyjth: Yep - since the behavior of the script has changed. you used to need to run it from the root

rkdavidson: Updated that, and removed all the whitespace changes my IDE decided were important :).

redreceipt (approved):

PR #413, opened 2021-06-24
locks down node version

redreceipt:
were having podfile lock changes on build, I'm guessing it's the node version but I haven't confirmed

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,6 +42,8 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - uses: actions/setup-node@v2
+      with:
+        node-version: 12
 
     - name: Install Dependencies
       run: yarn --ignore-scripts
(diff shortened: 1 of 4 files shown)


vinnyjth:
@redreceipt It's because we are updating the `npm` version of our media player package, which bumps the `pod` version of that package. Except we aren't capturing the `podfile.lock` changes w/ our `bump` tool.

redreceipt:
I recently updated the podfile lock https://github.com/ApollosProject/apollos-templates/commit/257452ed3cb5f5d861aafde8204e5fbc82ef1e18 and it didn't help

redreceipt:
our bump tool does run `yarn` which includes `yarn pods`

redreceipt:
https://github.com/ApollosProject/apollos-templates/blob/42224bd8ea83de16915a82faa9ec2e968b9884d6/scripts/bump.sh#L29

vinnyjth:
Weird, I see what you're saying and you're correct. And it's the GLOG version that's changing. Merging this seems like a good move.

vinnyjth (approved):

PR #415, opened 2021-06-25
make tabs a stack nav

redreceipt:
technically it doesn't matter because we are "switching" to it so you don't see the modal. but this is the right thing to do.


redreceipt:
don't _really_ need this one now that I switched the `reset` but it can't hurt

conrad-vanl (approved):

PR #416, opened 2021-06-26
uses new useLogout hook

redreceipt:
avoid deprecated local resolvers by using new hook that is client and navigation aware


redreceipt:
this will go in apps

PR #418, opened 2021-06-30
Remove ActionContainer

conrad-vanl:
Pairs nicely with https://github.com/ApollosProject/apollos-apps/pull/1838


redreceipt:
niiiice. would you rather there be a potential gap of time that has both bars or neither bars?

redreceipt:
actually, since we'll merge the other first, there's already potential to have both bars if they don't fix templates. So I'm gonna make this one a draft and we'll merge once the other is in a release so there's not a gap of time without either.

redreceipt (approved):

redreceipt:
just waiting on a core release at this point

PR #420, opened 2021-07-01
Update README.md

nlewis84:


redreceipt (commented):
I don't really wanna add this because those other docs are optional and not required for a working server or app. It's just what we do for our Di managed apps. I mention above in the doc the checkout the docs for more integrations.

PR #425, opened 2021-07-12
content-item-datasource companion PR

vinnyjth:


apolloschurchapp/ios/Podfile.lock
```diff
@@ -613,7 +613,7 @@ SPEC CHECKSUMS:
   boost-for-react-native: 39c7adb57c4e60d6c5479dd8623128eb5b3f0f2c
   BVLinearGradient: e3aad03778a456d77928f594a649e96995f1c872
   CocoaAsyncSocket: 065fd1e645c7abab64f7a6a2007a48038fdc6a99
-  DoubleConversion: cf9b38bf0b2d048436d9a82ad2abe1404f11e7de
+  DoubleConversion: cde416483dac037923206447da6e1454df403714
```
redreceipt, line 616: I've got to figure this out for myself...

apollos-church-api/config.yml
```diff
@@ -117,23 +95,23 @@ TABS:
...
-            - 9
+            - 0a269d99-2e1a-4fd3-aedf-7e0abc10e948
       type: HeroList
       isFeatured: true
       subtitle: Featured
     - title: FOR YOU
       algorithms: [SERMON_CHILDREN, PERSONA_FEED]
       subtitle: Explore what God calls you to today
       type: ActionList
-    - algorithms: [PERSONA_FEED]
```
redreceipt, line 136: I checked the `personaFeed` query and it seems ok, is it broken?

vinnyjth: I think I was frustrated when testing that there were two persona feeds, so I removed one of them :) we can put it back

redreceipt: nah, I'm good leaving like this, no redundancy. just wanted to make sure it wasn't broken

vinnyjth:
@redreceipt this is ready to merge

redreceipt:
got a migration error, and also images and content aren't coming through

Screen Shot 2021-07-19 at 2 40 05 PM
Kapture 2021-07-19 at 14 41 02

vinnyjth:
@redreceipt the failing migration is a migration that existed way before this body of work :/ I would recommend wiping out your postgres development server and starting from scratch. 

I'll preempt your question and argue that this is okay ;) 

If it works from scratch, that will match where most clients are starting from. All of our clients that I checked (newspring and rivervalley) haven't run that migration yet, but also don't have the `personId` field. They will not run into issues. 

As for
> images and content aren't coming through

You are pointing at your local postgres right? You'll also need to run the shovel locally. Or you can swap out your database url for the production postgres URL in the shared.env.enc

redreceipt:
I was pointing at the core production DB for these errors

vinnyjth:
@redreceipt Ah - that makes more sense actually. The DB was in a weird state since we were deving with it. I reset it and you should be gtg now.

vinnyjth:
image

redreceipt:
it's kicking me out immediately after login

Screen Shot 2021-07-19 at 3 10 39 PM


Kapture 2021-07-19 at 15 11 26

vinnyjth:
@redreceipt My bad again 😬 I didn't backfill people after I wiped the DB, give it another go. No need to change again, just try logging in again.

redreceipt:
closer
- the continue feature on the home screen is broken
- event images aren't coming through
- nothing is on the read tab
- the watch tab is broken

Screen Shot 2021-07-19 at 3 22 31 PM
Screen Shot 2021-07-19 at 3 22 35 PM
Screen Shot 2021-07-19 at 3 22 42 PM
Screen Shot 2021-07-19 at 3 22 44 PM

vinnyjth:
haha - I broke it when I wiped the DB (since the config.yml points at IDS in the DB now) 

Let me fix and confirm and then I'll circle back

vinnyjth:
@redreceipt Ready for review again. A few notes...

- Event images are broken, I'm considering that a won't fix for this PR. We will handle that by shoveling events in the future. 
- ButtonLink feature is broken. I want to merge this without a working ButtonLink feature, but will be working on a fast follow with the fix.

vinnyjth:
Continue feature being broken ... that's a tough one :/ we'll likely do a follow up PR that fixes that specific case. We will be able to come up with a really robust fix when we port in Interactions into postgres.

redreceipt:
So safe to assume churches that use Rock events (Willow may be the only one???) can't use the DB until we fix events?

vinnyjth:
> So safe to assume churches that use Rock events (Willow may be the only one???) can't use the DB until we fix events?

Correct! Well actually Willow might be fine since their implementation of events is largely custom anyway. 

I've captured the bugs here in this todo list: https://3.basecamp.com/3926363/buckets/6947926/todolists/3971672104

I really want to get this PR merged so we have some solid ground to work off of, once this is merged we'll be able to go back to smaller focused PR's instead of trying to work around the monolith :)

redreceipt:
I think it's important we treat this template as a working app with working features. The only reason I'm being a stickler is with more and more apps, the upgrade process needs to be smooth and with some features not working on the DB in the template, that could get very difficult to track. I'm committing to upgrading and releasing an app every eight weeks and need to be able to count on the template app not having accepted bugs in it.

redreceipt:
we decided to fix the **obvious** bugs before merging this PR, more updates to come...

vinnyjth:
Summary of slack huddle: 

Setting this to draft
Fixing bugs in this
When no obvious bugs are remaining, we look at a way to push this PR out in a way that doesn't inhibit our ability to update non content-item Postgres apps

vinnyjth:
@redreceipt  "Series in Progress" empty state is broken on master as well

image


It's actually a bug with the content card. The API is sending over the same response every time, but the front end doesn't render the message type correctly I think? 

```
{
  "data": {
    "node": {
      "id": "HorizontalCardListFeature:9724110384a5109532f35a4980ea9b7710c5d31e1ac7d657197c0b74fd401df69286cd1d18f9eee3d1c46520b8c4da91fa5f2d1ff423886173e0bdb7d612c7d714e28f3cb6472cdce5ea3ac72c4f6a35ec354b9d680a119775a33f4969995f3f07a8d14629ff99d33c7651baccf3e8f34f382889a21013916015f009776eeb84",
      "title": null,
      "subtitle": "Continue",
      "cards": [
        {
          "action": null,
          "title": null,
          "hyphenatedTitle": null,
          "hasAction": null,
          "actionIcon": null,
          "labelText": null,
          "summary": null,
          "coverImage": null,
          "relatedNode": {
            "id": "Message:99511d5ba7917b86d61382a46ad38396",
            "message": "All caught up!",
            "__typename": "Message"
          },
          "__typename": "CardListItem"
        }
      ],
      "primaryAction": null,
      "__typename": "HorizontalCardListFeature"
    }
  }
}
```

redreceipt:
well dang. I hated that feature as I was writing it...

vinnyjth:
@redreceipt Once this release finishes (https://github.com/ApollosProject/apollos-apps/actions/runs/1050109938) this PR will be ready for review again. Taking it back out of draft.

redreceipt:
error when you're not using PG (`DATABASE_URL` is blank) 

Screen Shot 2021-07-21 at 11 55 38 AM

vinnyjth:
@redreceipt yahhh this is going to be problamatic. The old config.yml and the new postgres config.yml use different formats. Should we create a `config.postgres.yml`?

vinnyjth:
@redreceipt actually I can't reproduce locally, are you getting an error in your console?

redreceipt:
no it's a network error, I'll track it down. you probably can't reproduce unless you log out in between. i bet it's the config yml thing

vinnyjth:
@redreceipt I can reproduce, and it looks like an error in Rock caused by trying to filter by a `guid` coming from the config.yml

redreceipt (approved):

PR #430, opened 2021-07-14
removes events

redreceipt:
Screen Shot 2021-08-17 at 4 58 17 PM


apolloschurchapp/src/index.js
```diff
@@ -109,11 +108,6 @@ const App = () => (
                 stackPresentation: 'push',
               })}
             />
-            <Screen
-              name="Event"
```
conrad-vanl, line 113: should we put in a redirect from `Event` to content? I know if you try to navigate to a route that doesn't exist, you'll get an app crash

redreceipt:
Why/how would people navigate to events?

conrad-vanl:
They had an event opened before they upgraded to this version of the app
The API had an event as the action to link to (like in a feature) and either the app had cached content or the api wasn't properly updated
A push notification was sent out with /event in the url

I'm happy approving without this change....I was hoping react-navigation had an easy way to do redirects......

redreceipt:
Oh yea valid. I think those are pretty rare cases but could happen maybe, I’d rather figure out how to silently fail.

redreceipt:
@conrad-vanl react-navigation 6 has built-in deep link support complete with a catch-all that we should probably move to. But I don't it's worth waiting on this PR considering I believe Willow is the only app that uses official events.

conrad-vanl (approved):

PR #432, opened 2021-07-15
changes to new theme provider

redreceipt:


apolloschurchapp/src/index.js
```diff
@@ -12,12 +13,14 @@ import { createNativeStackNavigator } from 'react-native-screens/native-stack';
...
 import 'react-native-gesture-handler'; // required for react-navigation
 import { enableScreens } from 'react-native-screens';
+import { SafeAreaProvider } from 'react-native-safe-area-context';
+import { BottomSheetModalProvider } from '@gorhom/bottom-sheet';
 
 import {
   BackgroundView,
-  withTheme,
   NavigationService,
-  Providers as ThemeProvider,
```
conrad-vanl, line 20: I'm pretty sure we just need to add `Themer` to the[ list of providers](https://github.com/ApollosProject/apollos-apps/blob/master/packages/apollos-ui-kit/src/Providers.js#L8) that `ui-kit` exports. Doing it this way I think we'll lose some of the [other providers ](https://github.com/ApollosProject/apollos-apps/blob/master/packages/apollos-ui-kit/src/Providers.js#L8)in ui-kit?

apolloschurchapp/src/index.js
```diff
@@ -77,72 +97,79 @@ const LandingToAuth = () => {
...
-      <ThemedNavigationContainer
-        containerRef={NavigationService.setTopLevelNavigator}
-        onReady={NavigationService.setIsReady}
-      >
-        <Providers>
-          <Navigator
-            screenOptions={{ headerShown: false, stackPresentation: 'modal' }}
+  <Themer theme={customTheme} icons={customIcons}>
+    <SafeAreaProvider>
+      <BottomSheetModalProvider>
```
redreceipt, line 102: @conrad-vanl we won't lose them, I added them here. I think it's better because it's more explicit and you know what the various contexts are. Plus it avoids `recompose.nest`

conrad-vanl: Yeah but this adds _more_ code to the template project, getting further away from template 0 right? Like now if we add a new provider, all of our frontend apps have to be updated. We can get rid of `nest()` inside of ui-kit if that's a concern.

redreceipt: I personally think it's a bad abstraction for us to export another package's context provider. Likely _all_ the providers in the template will eventually be exported from NPM, but being explicit here is a better way to do it.

redreceipt: for more context here our `LayoutProvider` is simply `SafeAreaProvider` re-exported

conrad-vanl: > I personally think it's a bad abstraction for us to export another package's context provider.

Why?

currently, this is what ui-kit does, and to use ui-kit as that package is intended, we should import the `Provider` from it and render it here, not render out the providers separately. I don't think I'm comfortable approving splitting out to separate Providers unless we decide to change how UI-Kit works.

And still, I still argue that this is additional code and maintenance needed. If ui-kit adds another provider that is needed, you have to modify each app's code to incorporate that provider. This requires additional code ownership on each app.... I think that's the primary argument for the abstraction.

conrad-vanl (changes requested):

redreceipt:
I also trying to avoid exporting a `NewProviders` from `ui-kit` that takes out the old `ThemeProvider`

redreceipt:
I think eventually _all_ providers will be exported, analytics, live, everything, from one place. probably the top-level `ApollosApp` export. Just don't think we need the exports also coming from `ui-kit`

apolloschurchapp/src/index.js
redreceipt: Would you prefer I export a new `UIProviders` component from `ui-kit` without the old `ThemeProvider`?

redreceipt:
ok @conrad-vanl I've got a PR that preserves `Providers` 

https://github.com/ApollosProject/apollos-apps/pull/2156

PR #434, opened 2021-07-17
replaces config variables with env variables

redreceipt:


conrad-vanl (commented):
I'm confused how this works?

redreceipt:
> I'm confused how this works?

This is the draft PR that's tied to this: https://github.com/ApollosProject/apollos-apps/pull/2067

PR #442, opened 2021-08-03
add certs lane to fastlane

redreceipt:
lets us create certs and push certs easily


conrad-vanl (commented):
Do we need to make a documentation update for this?

redreceipt:
not sure if this is the right answer but this is one of those that I wanted to test out and use before writing official documentation. wanna make sure it doesn't need to be tweaked a bit first. As is, it's completely opt-in, won't run by itself. But it will make it super simple for people to gen certificates and push cert for the first time.

conrad-vanl (approved):

PR #443, opened 2021-08-03
Node 14 / Use External GraphQL Upload Package

vinnyjth:


apollos-church-api/package.json
```diff
@@ -14,7 +14,7 @@
     "translate": "node scripts/translate.js"
   },
   "engines": {
-    "node": "12"
+    "node": "14"
```
redreceipt, line 17:
```suggestion
    "node": ">=14"
```

apolloschurchapp/package.json
```diff
@@ -3,7 +3,7 @@
   "version": "1.0.0",
   "private": true,
   "engines": {
-    "node": "12"
+    "node": "14"
```
redreceipt, line 6:
```suggestion
    "node": ">=14"
```

apollos-church-api/package.json
vinnyjth: Good call.

.github/workflows/ci.yml
```diff
@@ -10,7 +10,7 @@ jobs:
     - uses: actions/checkout@v2
     - uses: actions/setup-node@v2
       with:
```
redreceipt, line 12: I like removing all these "with" lines so it uses the current version which will let us know when there's new version of node that isn't working with our project

vinnyjth: If we go that route, I would rather test with multiple versions of Node so we can control our environment and test for node version specific failures.

redreceipt: I guess I don't see the point in knowing about failures from different versions. It would only fail once and then we'd fix it. Pointing to the latest will automatically let us know when a new version of `node` is mainstream and doesn't work with the project. Currently, we don't really have a good way of being notified about that.

vinnyjth: I don't want to end up in the world where Github updates their node version and our builds start breaking (with only the ability to guess about the reason of the failure). Testing with multiple versions will at least grant us quick insight into the nature of a failure if one version remains passing while a newer version starts failing.

vinnyjth: What do you think about leaving this PR the way it is for now, given that it matches the previous behavior of specifying a node version, and move the decision making about how to handle testing for different node versions somewhere else?

redreceipt: Shoot sorry I thought I responded to this. Yea I agree, let's leave it.

redreceipt (approved):

PR #445, opened 2021-08-05
lock down in-app browser version

redreceipt:
3.6.3 is broken, tested on core. new apps are jumping to the broken version


redreceipt:
this is breaking android builds

conrad-vanl (approved):

PR #447, opened 2021-08-11
dont need graphql-upload pkg as a dependency

redreceipt:


redreceipt:
cant test right now, getting kicked out of the core app template

conrad-vanl (approved):

PR #449, opened 2021-08-12
opt in to content in the database

redreceipt:
the migration is not trivial so in the meantime, we need a path for a smooth upgrade process. this makes database content opt-in


apollos-church-api/src/config.js
```diff
@@ -3,9 +3,10 @@ import fetch from 'node-fetch';
 import dotenv from "dotenv/config"; // eslint-disable-line
 import ApollosConfig from '@apollosproject/config';
 
-const configName = process.env.DATABASE_URL
-  ? 'config.postgres.yml'
-  : 'config.yml';
+const configName =
+  process.env.DATABASE_URL && process.env.DATABASE_CONTENT === 'true'
```
conrad-vanl, line 7: Instead of basing off of a env variable name (which is a bit of technical debt), could we instead just do it off the existing of the file? Try to load `config.postgres.yml` and if that fails, fallback to `config.yml`?

redreceipt: I like that a lot

redreceipt: Oh another thing, we have to protect against certain imports in the `server/data/index.postgres.js` file so I guess I can check for the existence of the file in there too

redreceipt:
ok, @conrad-vanl @vinnyjth I think this will work nicely for us. It allows us to seamlessly switch (and upgrade) with Rock or Postgres content just by the presence of the `config.postgres.yml` file. I tested but feel free to give it a spin.

conrad-vanl (approved):

PR #452, opened 2021-08-17
chore: Stop running migrations automatically, run migrations by hand. 

vinnyjth:
Steps to test
1. Make sure you are pointed at your local database schema (not production) 
2. From the API folder, run `yarn migrator down` 
3. Boot the API, look for a special message telling you that you have migrations to run
4. Run `yarn migrator up`. 
5 Confirm that migrations were ran. 

Also note commands like `yarn migrator --help`


nlewis84:
Could we put in some regex that would require the database URL variable to start with "http://localhost:"? And throw a warning to switch the variable.

apollos-church-api/package.json
```diff
@@ -5,6 +5,8 @@
   "main": "lib/index.js",
   "scripts": {
     "setup": "./scripts/init.sh",
+    "migrator": "node ./lib/migrator.js",
+    "migrator:production": "heroku --app YOUR_HEROKU_APP_NAME_HERE run yarn migrator up",
```
redreceipt, line 9: Could we make migrations automatic on the Heroku app? Is that not the purpose of the `AUTO_MIGRATE` variable?  I just hate introducing named specific things to the template that don't work out of the box and require more documentation and setup.

vinnyjth: I don't think migrations should be automatic ever. Except for "power users" who want to run migrations automatically to make development easier. I'm fine deleting the coded that let's migrations run automatically.

redreceipt:
> Could we put in some regex that would require the database URL variable to start with "http://localhost:"? And throw a warning to switch the variable.

don't love that because we may use copies or staging DBs at some point. it's really just protecting the prod DB we're concerned about.

apollos-church-api/package.json
redreceipt: what about just leaving this line off then? I like the shortcut but I don't like the specific name and setup required even more. I feel like someone should have to ask and know what they're doing before running migrations in production. having this line in the setup readme should suffice.

vinnyjth: fine by me ;) I had assumed that if I left this line out you were going to ask me to add it in

vinnyjth:
```suggestion
```

vinnyjth:
@redreceipt  added some documentation to the readme

README.md
```diff
@@ -82,6 +82,20 @@ heroku open
 
 To get started with different API integrations, check out our [docs](https://apollosapp.io)!
 
+#### 
+rations
```
redreceipt, line 86:
```suggestion
#### Migrations
```

README.md
```diff
@@ -82,8 +82,7 @@ heroku open
 
 To get started with different API integrations, check out our [docs](https://apollosapp.io)!
 
-####
```
vinnyjth: agh you beat me to it!

redreceipt (approved):

PR #453, opened 2021-08-17
properly kick people out for authentication errors

redreceipt:
BEFORE

Kapture 2021-08-17 at 15 09 28

AFTER

Kapture 2021-08-17 at 15 13 22


redreceipt:
the only thing that makes me nervous about this PR is there has to be a reason we didn't have it `client.clearStore()` from the start and needed the hack...

vinnyjth (approved):

PR #454, opened 2021-08-17
safer database content opt in

redreceipt:
the other way doesn't work like i'd hoped with upgrades because the config file is patched from templates and then its confusing why the app isn't working. this is just temporary, soon all apps will use database content anyway, and we won't need this opt in. maybe can drop it for v3 when DB is required.


vinnyjth:
I was afraid that would be an issue :/

vinnyjth (approved):

PR #457, opened 2021-08-18
ios metadata

redreceipt:


conrad-vanl (approved):
Nice! How much of this is generated when you generate icons and such? Could these images be generated? Should we add filling out this metadata to the instructions/readme?

redreceipt:
these images can be generated but it involves swift and the TESTS xcode project and seems very complex. maybe for a future PR.

PR #460, opened 2021-08-20
changed match auth variable name and more explicit gh secret script

nlewis84:


nlewis84:
The spaces got added by prettier. Also, gonna update a picture in here.

redreceipt (approved):

PR #463, opened 2021-08-23
adds custom theme docs

redreceipt:
this also removes the placeholder files from the template


redreceipt:
@conrad-vanl could we use a built-in font for the default? I've written instructions how to customize. And it will help us get to template zero

conrad-vanl:
> @conrad-vanl could we use a built-in font for the default?

I'm currently not in favor of dropping Inter.

But, perhaps we can figure out a way to have Inter installed via our ui-kit package? I know that react-native NPM dependencies can add other asset types...

redreceipt:
sounds good, that can be for a different PR, this one doesn't touch the fonts yet

conrad-vanl (changes requested):
Since modifying the theme is something every app does, I think we should keep a skeleton file with that ready to go. Similar to how we have CONFIG.yaml inside of -api. That way we don't have to jump in and create the theme.js file and such with every app we spin up.

redreceipt:
shouldn't documentation serve that purpose if we're trying to get to template zero? especially since an actual file doesn't really matter, rather just passing in the object to the `Themer` props?

vinnyjth:
@redreceipt My $0.02 is an empty file with code commented out is better than documentation since it's that much easier to comment code in, than find the documentation in the first place.

conrad-vanl:
Doing it this way I think we're going to create discrepancies on how we "override" theme from one app to the next. Maybe if we only have one or two overrides, we'll just pass an object into `<Themer>` like you say, but another app might have a lot of overrides and we decide to split out into a separate file. That creates increased technical debt and ownership cost...

Template-Zero to me is about getting to a point where the only files that are left in the "template" are files that you are expected to customize. That might end up being a single file that looks like:

```
import App from '@apollosproject/ui-bootstrap-react-native';

const RiverValley = () => (
  <App
    apiEndpoint={'https://rivervalley.herokuapp.com'}
    theme={{
       colors: { ... },
    }}
    localization={{
       en: {
          userMessagesUnread: () => '...',
       },
    }}
  />
);

export default RiverValley;
```

Or perhaps:

```
import App from '@apollosproject/ui-bootstrap-react-native';

const RiverValley = () => (
  <App
    apiEndpoint={'https://rivervalley.herokuapp.com'}
    theme={require('./theme.js')}
    localization={require('./localization.js')}
  />
);

export default RiverValley;
```

Or even:

```
import App from '@apollosproject/ui-bootstrap-react-native';

const RiverValley = () => (
  <App appKey="rivervalley" > {/* configuration downloaded over the wire from multi-tenant apollos api server */}
);

export default RiverValley;
```

Which would be great! But nonetheless within any model we choose, I think we need to have only _one_ way that we provide customization, and that customization should probably be stubbed out in the template to the point that any boilerplate code is already established within the template. When we start copy-pasting boilerplate code from readmes I think that opens us up into too much potential technical debt.

redreceipt:
I think I'm trying to move away from a generated skeleton project in favor of building an Apollos app from nothing using libraries. I would argue good APIs that are hard to mess up don't result in technical debt and are better than boilerplate code. But I'll put this on the shelf, it's definitely aggressive and ahead of our current needs.

conrad-vanl:
@redreceipt I would say most of what's in this PR is great...and I don't think what you are saying and what @vinnyjth and I are saying are mutually exclusive, I just think they speak to two separate perspectives.

The perspective I'm concerned with is as a team, we have to manage many very similar apps with separate code repos. Reducing the amount of code in each of those repos is important, but so is making sure that each of those code repos looks, smells, etc the same are equally important. My concern with the current approach in this PR is it leaves implementation of something like how we override theme and where up to the developer, and that can create a consistency issue, which causes code debt. Even something as simple as doing something like `<Themer theme={{ colors: { ... }} />` and `const theme = {...}; <Themer theme={theme} />` between apps is unnecessary code debt. 

The perspective I think you are speaking to is still equally important and I agree 100%. I still think the amount of boilerplate we have, even with how theme is boilerplated, is too much.

PR #465, opened 2021-08-23
removes wordmark by default

redreceipt:
removing this from the default because it's very straigtforward to add back and it's something that looks fine with or without


conrad-vanl:
What if we put the logo source files into the theme? And then you could get to template zero by porting this code over into -apps?

redreceipt:
so export jpgs by default?

conrad-vanl:
The theme would just have a value like:

```
brand: {
  fullLogo: require('../path/to/logo'),
  ...anyOtherBrandImages
}
```

and the default `brand` would use the Apollos Logo

vinnyjth:
I like what @conrad-vanl is suggesting

redreceipt:
What if the default was just an SVG exported from icons and it worked exactly like the `BrandIcon` on the landing page?

conrad-vanl:
I don't think I'm opposed to that. Only problem with our Icons is that they're single color only, and I think most of our church logos are multi color. Not all of them though....it would make dark/light mode support easier.

redreceipt:
Oh, I didn't think about colors. I was mainly challenging the notion that it needs to be in there at all by default. But I'll table this for now.

nlewis84:
VCC has a multi color wordmark that stays the same for dark and light mode.

PR #472, opened 2021-09-01
fix: Updated Connect tab

nlewis84:
This is in preparation for moving the User Settings to an avatar in the header of all tabs.

Reliant on these two PRs:

https://github.com/ApollosProject/apollos-templates/pull/365

https://github.com/ApollosProject/apollos-apps/pull/2165


redreceipt:
@nlewis84 this still relevant? those files are gone now

nlewis84:
No, I don't think this is relevant any more with those files gone.

PR #475, opened 2021-09-07
chores: create shovel setup docs

tomthornton:


redreceipt (approved):

redreceipt:
@nlewis84 once this is merged, try to follow it exactly to set up vineyard, editing along the way as you hit bumps and get questions answered

PR #476, opened 2021-09-09
fix: Item liking sans content

vinnyjth:
This bridge code was only ever intended to be used w/ churches that used Rock content prior to migrating the Following feature to postgres. Now that we have migrated, it doesn't need to exist. Churches that use `DATABASE_CONTENT=true` should use the new `Like` dataSource, and the older churches should continue to use `Following`. 

Proof code works w/ Rock

Proof code works w/ Postgres


redreceipt (approved):
You know I'll blindly approve any PR that deletes a bunch of stuff!

conrad-vanl:
Need to get this implemented on FNWA asap. I can handle Chase Oaks.

redreceipt:
I’ll throw it in fellowship

PR #479, opened 2021-09-13
fix: Update Universal Links Docs and Template

nlewis84:
Set a new default value for associated domains in Xcode and added testing documentation for Universal Links.

### DIFF
--- a/apolloschurchapp/ios/apolloschurchapp/apolloschurchapp.entitlements
+++ b/apolloschurchapp/ios/apolloschurchapp/apolloschurchapp.entitlements
@@ -6,7 +6,7 @@
 	<string>development</string>
 	<key>com.apple.developer.associated-domains</key>
 	<array>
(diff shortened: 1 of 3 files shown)


apolloschurchapp/ios/apolloschurchapp/apolloschurchapp.entitlements
```diff
@@ -6,7 +6,7 @@
 	<string>development</string>
 	<key>com.apple.developer.associated-domains</key>
 	<array>
-		<string>applink:$(APP_DATA_URL)</string>
+		<string>applinks:apollos-church-api.herokuapp.com</string>
```
redreceipt, line 9: I don't like this because if we forget to change it (or someone in the open-source community forgets) it's wrong and would cause problems. I'd rather leave it off and add it to the docs unless we can figure out another way to make it universal

redreceipt (approved):

redreceipt (approved):

PR #480, opened 2021-09-14
adds android app ID to init script

redreceipt:


nlewis84 (approved):
Looks good to me and I tested it and it works as expected.

PR #491, opened 2021-09-28
all tabs with avatars

redreceipt:
Screen Shot 2021-09-28 at 11 13 43 AM

### DIFF
--- a/apolloschurchapp/src/tabs/index.js
+++ b/apolloschurchapp/src/tabs/index.js
@@ -134,7 +134,7 @@ const CustomConnectScreen = () => (
 const ConnectTabStack = createNativeStackNavigator();
 const ConnectTabStackNavigator = () => (
   <ConnectTabStack.Navigator
-    screenOptions={{
+    options={{
       headerHideShadow: true,
       headerLargeTitle: true,
       headerLeft: HeaderLeft,
--- a/apolloschurchapp/src/tabs/index.js
+++ b/apolloschurchapp/src/tabs/index.js
@@ -120,7 +120,7 @@ const WatchTab = createFeatureFeedTab({
 });
 
 const PrayTab = createFeatureFeedTab({
-  screenOptions: {
+  options: {
     headerLeft: HeaderLeft,
   },
   tabName: 'Pray',
--- a/apolloschurchapp/src/tabs/index.js
+++ b/apolloschurchapp/src/tabs/index.js
@@ -135,7 +135,6 @@ const ConnectTabStack = createNativeStackNavigator();
 const ConnectTabStackNavigator = () => (
   <ConnectTabStack.Navigator
     options={{
-      headerHideShadow: true,
       headerLargeTitle: true,
       headerLeft: HeaderLeft,
     }}
--- a/apolloschurchapp/src/tabs/index.js
+++ b/apolloschurchapp/src/tabs/index.js
@@ -134,13 +134,18 @@ const CustomConnectScreen = () => (
 const ConnectTabStack = createNativeStackNavigator();
 const ConnectTabStackNavigator = () => (
   <ConnectTabStack.Navigator
-    options={{
+    screenOptions={{
       headerHideShadow: true,
       headerLargeTitle: true,
-      headerLeft: HeaderLeft,
     }}
   >
-    <ConnectTabStack.Screen name={'Connect'} component={CustomConnectScreen} />
+    <ConnectTabStack.Screen
+      name={'Connect'}
+      component={CustomConnectScreen}
+      options={{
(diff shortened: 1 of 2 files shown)


apolloschurchapp/src/tabs/index.js
```diff
@@ -111,20 +104,46 @@ const HomeTab = createFeatureFeedTab({
 });
 
 const ReadTab = createFeatureFeedTab({
+  screenOptions: {
```
conrad-vanl, line 107: So this is a small thing, but this should be:

```suggestion
  options: {
```

The difference is that `screenOptions` seems to override not only this screen, but any children. So like if you have a "View All" button on the tab that opens up a nested route - the user profile avatar is still visible. Using `options` means that it's only applied to the header on this tab, and if you navigate to a child route you'll see a back button!

I did it this way [on Chase Oaks](https://github.com/Differential/chase-oaks-apollos/blob/master/apolloschurchapp/src/tabs/index.js#L96) after experimenting with both.

apolloschurchapp/src/tabs/index.js
```diff
@@ -111,20 +104,46 @@ const HomeTab = createFeatureFeedTab({
...
+    headerLeft: HeaderLeft,
+  },
   tabName: 'Read',
   feedName: 'READ',
 });
 
 const WatchTab = createFeatureFeedTab({
+  screenOptions: {
```
conrad-vanl, line 115:
```suggestion
  options: {
```

apolloschurchapp/src/tabs/index.js
```diff
@@ -111,20 +104,46 @@ const HomeTab = createFeatureFeedTab({
...
+    headerLeft: HeaderLeft,
+  },
   tabName: 'Watch',
   feedName: 'WATCH',
 });
 
 const PrayTab = createFeatureFeedTab({
+  screenOptions: {
```
conrad-vanl, line 123:
```suggestion
  options: {
```

apolloschurchapp/src/tabs/index.js
```diff
@@ -111,20 +104,46 @@ const HomeTab = createFeatureFeedTab({
...
 });
 
+const CustomConnectScreen = () => (
+  <ConnectScreenConnected ActionTable={ActionTable} ActionBar={ActionBar} />
+);
+
+const ConnectTabStack = createNativeStackNavigator();
+const ConnectTabStackNavigator = () => (
+  <ConnectTabStack.Navigator
+    screenOptions={{
```
conrad-vanl, line 137:
```suggestion
    options={{
```

apolloschurchapp/src/tabs/index.js
```diff
@@ -111,20 +104,46 @@ const HomeTab = createFeatureFeedTab({
...
+      headerHideShadow: true,
```
conrad-vanl, line 138:
```suggestion
```

conrad-vanl (changes requested):
Thoughts...looks good otherwise!

apolloschurchapp/src/tabs/index.js
conrad-vanl: (don't think we need to hide the shadow just on this tab? We don't on any other tab)

apolloschurchapp/src/tabs/index.js
redreceipt: oh nice

apolloschurchapp/src/tabs/index.js
```diff
@@ -111,20 +104,45 @@ const HomeTab = createFeatureFeedTab({
...
 
+const CustomConnectScreen = () => (
+  <ConnectScreenConnected ActionTable={ActionTable} ActionBar={ActionBar} />
+);
+
+const ConnectTabStack = createNativeStackNavigator();
+const ConnectTabStackNavigator = () => (
+  <ConnectTabStack.Navigator
+    options={{
+      headerLargeTitle: true,
```
redreceipt, line 138:
```suggestion
      headerHideShadow: true,
      headerLargeTitle: true,
```

redreceipt: It actually is on the `ui-connected` component. there is a chance we don't need it since it's a different color but having this for now matches

redreceipt:
ok @conrad-vanl ready now

apolloschurchapp/src/tabs/index.js
redreceipt: Never mind I was confused. Point is we are actually hiding shadows. Maybe we don't need to be and it would look ok but that can be for another PR.

conrad-vanl (approved):

PR #498, opened 2021-10-21
feat: Following screen

tomthornton:
Related to this PR
https://github.com/ApollosProject/apollos-apps/pull/2217


tomthornton:
@redreceipt that's true. It seems like the connect tab is going to be phased out at some point. @conrad-vanl what do you think?

